PR #1, opened 2022-01-03
fix Cursor::next_line_remaining() with non-ascii characters

koineisation:
`Cursor::next_line_remaining()` would fail if the remaining line has any non-ascii character, since the original code assumes `self.source[self.index..self.index + 1]` to be a valid utf-8 string slice.

I don't know if I've done right with `self.col` and `self.index`. Please fix it if I got it wrong.


MikuroXina:
Thank you for contributing! I found that operating the non-ASCII characters with `Cursor` got wrong about numbers of columns control. I will fix about them.

PR #35, opened 2023-10-04
Fix Bgm Obj can't be fully parsed

kznr02:
Before Fix:
image

After Fix:
image

### DIFF
--- a/src/parse/notes.rs
+++ b/src/parse/notes.rs
@@ -386,10 +386,7 @@ impl Notes {
                 message,
             } => {
                 for (time, obj) in ids_from_message(*track, message) {
-                    self.bgms
-                        .entry(time)
-                        .and_modify(|vec| vec.push(obj))
-                        .or_insert_with(Vec::new);
+                    self.bgms.entry(time).or_default().push(obj)
                 }
             }
             Token::Message {


src/parse/notes.rs
```diff
@@ -389,7 +389,7 @@ impl Notes {
                     self.bgms
                         .entry(time)
                         .and_modify(|vec| vec.push(obj))
-                        .or_insert_with(Vec::new);
+                        .or_insert_with(|| vec![obj]);
```
MikuroXina, line 392: Thank you for contribution! But this fix could be written more cleanly like:
```rs
self.bgms.entry(time).or_default().push(obj);
```

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bms-rs"
-version = "0.4.1"
+version = "0.4.2"
```
MikuroXina, line 3: Please don't bump the version manually, because it will be updated automatically with CI.

tests/bgm.rs
```diff
@@ -0,0 +1,12 @@
...
+#[test]
+fn bgm_obj_test() {
+    let src = std::fs::read_to_string("tests/nc_mx.bme").unwrap();
+    let ts = parse(&src).unwrap();
+    let rng = RngMock([1]);
+    let bms = Bms::from_token_stream(&ts, rng).unwrap();
+    for b in bms.notes.bgms() {
+        println!("{:?}", b);
+    }
+}
```
MikuroXina, line 12: I think that this test is doing same thing as `test_nc` in `tests/files.rs`, so it should not be added.

MikuroXina (changes requested):

kznr02:
So sorry about my mistakes ><, I'll more careful next time. Thanks for your correction, I have corrected the code.

MikuroXina (approved):
No problem, I'm glad to your contribution!

PR #37, opened 2023-10-05
Fix non-standard resource name can't be handled correctly.

kznr02:
1. Fix the wav file with blank space can't be fully parse. (Freedom Dive)
2. Incorrect resource file won't be parsed. (#WAV/#BMP filename which is without prefix/extension or just a blank name)

### DIFF
--- a/src/lex/cursor.rs
+++ b/src/lex/cursor.rs
@@ -67,11 +67,14 @@ impl<'a> Cursor<'a> {
     }
 
     pub(crate) fn next_line_remaining(&mut self) -> &'a str {
+        fn is_blank_or_tab(c: char) -> bool {
+            return c == ' ' || c == '\t';
+        }
         let spaces = self.source[self.index..]
-            .find(|c: char| !c.is_whitespace())
+            .find(|c: char| !is_blank_or_tab(c))
             .unwrap_or(self.source.len());
 
-        self.col += spaces;
+        self.col += 1;
         self.index += spaces;
 
         let remaining_end = self.source[self.index..]
--- a/src/lex/cursor.rs
+++ b/src/lex/cursor.rs
@@ -67,16 +67,6 @@ impl<'a> Cursor<'a> {
     }
 
     pub(crate) fn next_line_remaining(&mut self) -> &'a str {
-        fn is_blank_or_tab(c: char) -> bool {
-            return c == ' ' || c == '\t';
-        }
-        let spaces = self.source[self.index..]
-            .find(|c: char| !is_blank_or_tab(c))
-            .unwrap_or(self.source.len());
-
-        self.col += 1;
-        self.index += spaces;
-
         let remaining_end = self.source[self.index..]
             .find('\n')
             .unwrap_or(self.source.len());
@@ -91,7 +81,7 @@ impl<'a> Cursor<'a> {
         };
         self.col += ret.chars().count();
         self.index += remaining_end;
-        ret
+        ret.trim()
     }
 
     pub(crate) fn line(&self) -> usize {
--- a/src/lex/token.rs
+++ b/src/lex/token.rs
@@ -241,16 +241,20 @@ impl<'a> Token<'a> {
                 }
                 wav if wav.starts_with("#WAV") => {
                     let id = command.trim_start_matches("#WAV");
+                    let str = c.next_line_remaining();
                     let filename = Path::new(
-                        c.next_token()
+                        (!(str.is_empty()))
+                            .then(|| str)
                             .ok_or_else(|| c.err_expected_token("key audio filename"))?,
                     );
                     Self::Wav(ObjId::from(id, c)?, filename)
                 }
                 bmp if bmp.starts_with("#BMP") => {
                     let id = command.trim_start_matches("#BMP");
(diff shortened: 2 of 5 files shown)


tests/dive_n7.bme
```diff
@@ -0,0 +1,4522 @@
...
+#12801:00
+#12801:00
+#12801:00
+#12801:00
+#12801:00
+#12801:9W
+#12804:34
+#12811:1F
+#12816:1N
+
```
MikuroXina, line 4522: It seems that this is an original one created by xi. Did you obtain permission to redistribute this file? All other BMS files uploaded here were created by me. I want to be clean about copyright, so you ought to obtain permission for commiting it.

Or instead you can add a test case about BMS that contians only `#WAV` and `#BMP` to resolve it.

MikuroXina (changes requested):

src/lex/cursor.rs
```diff
@@ -68,10 +68,10 @@ impl<'a> Cursor<'a> {
 
     pub(crate) fn next_line_remaining(&mut self) -> &'a str {
         let spaces = self.source[self.index..]
-            .find(|c: char| !c.is_whitespace())
+            .find(|c: char| c != ' ')
             .unwrap_or(self.source.len());
 
-        self.col += spaces;
+        self.col += 1;
```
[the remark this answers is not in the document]
kznr02: That's really a problem, many characters are defined as whitespace, so I'm looking for a solution to make this check as useful as possible. is_whitespace() method seems will drop some special contents.

src/lex/token.rs
```diff
@@ -241,18 +241,18 @@ impl<'a> Token<'a> {
...
                 }
                 bmp if bmp.starts_with("#BMP") => {
                     let id = command.trim_start_matches("#BMP");
-                    let filename = Path::new(
-                        c.next_token()
-                            .ok_or_else(|| c.err_expected_token("bgi image filename"))?,
-                    );
+                    let filename = Path::new(c.next_line_remaining());
+                    filename
+                        .extension()
```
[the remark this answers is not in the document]
kznr02: Perhaps just checking if this filename is blank is enough?

tests/dive_n7.bme
kznr02: I've never considered about it, thanks for noticing me so much, I will create a test document instead of uploading the original one.

src/lex/cursor.rs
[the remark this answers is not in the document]
MikuroXina: I found that the extraction process is too redundant, and it can be written as:
1. Get a slice of string from now to the next new line character.
2. Trim it whose start and end.
3. Seek to next line.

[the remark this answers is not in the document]
kznr02: Yes, I found remove the '!' of `!c.is_whitespace()` while leave the whitespace in final string, so just trim it and return it, that could be the better way and less redundant code.

src/lex/cursor.rs
```diff
@@ -88,7 +88,7 @@ impl<'a> Cursor<'a> {
         };
         self.col += ret.chars().count();
         self.index += remaining_end;
-        ret
+        ret.trim()
```
MikuroXina, line 91: That's good. For improvement, you can omit skipping whitespaces with the line 70-76 and 80-88. `ret.trim()` will remove them.

tests/files.rs
```diff
@@ -26,3 +26,15 @@ fn test_j219() {
     let bms = Bms::from_token_stream(&ts, RngMock([1])).expect("must be parsed");
     eprintln!("{:?}", bms);
 }
+
+#[test]
+fn test_blank() {
+    let source = include_str!("dive_withblank.bme");
+    let ts = parse(source);
+    assert_eq!(ts.is_err(), true);
```
MikuroXina, line 34: Hmm, I'd better to check whether the actual `Err` is an expected one. And you should remove these commented out codes.

MikuroXina (changes requested):

tests/files.rs
kznr02: That means the actual error type would be implemented `Eq` trait, I'm not sure if you want to do that, but I add it in the code.

src/lex/cursor.rs
kznr02: Lines 80-88 should be kept, the error message will include `col`, for accuracy this check is necessary.

src/lex/token.rs
```diff
@@ -241,16 +241,20 @@ impl<'a> Token<'a> {
                 }
                 wav if wav.starts_with("#WAV") => {
                     let id = command.trim_start_matches("#WAV");
+                    let str = c.next_line_remaining();
                     let filename = Path::new(
-                        c.next_token()
+                        (!(str.is_empty()))
+                            .then(|| str)
                             .ok_or_else(|| c.err_expected_token("key audio filename"))?,
                     );
```
MikuroXina, line 249:
```rs
if str.is_empty() {
    return Err(c.err_expected_token("key audio filename"));
}
let filename = Path::new(str);
```
Using a simple `if` could make it more readable. This method chain may be difficult to read.

MikuroXina (approved):

PR #89, opened 2024-12-28
Lex: Remove todo!() there to avoid unexpected behavior when meeting new command like #SCROLL which is not supported by this library.

MiyakoMeow:
`Lex`:
1. Remove `todo!()` there to avoid unexpected behavior when meeting new command like #SCROLL which is not supported by this library.
2. Split `LexWarning` from `LexError`.
3. Modified the behavior of `lex::parse()`, and renamed some elements. New function `lex::parse_and_get_warnings()` added, the original `lex::parse()` is there, but now that is not recommand. The usage of this library is not modified, when old codes relying it can be built with the action changed above.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -21,7 +21,7 @@ jobs:
             with:
                 app_id: ${{ secrets.HELPER_BOT_APP_ID }}
                 private_key: ${{ secrets.HELPER_BOT_PRIVATE_KEY }}
-        -   uses: google-github-actions/release-please-action@v4
+        -   uses: googleapis/release-please-action@v4
             with:
                 release-type: rust
                 token: ${{ steps.token.outputs.token }}


MikuroXina:
Thanks for contribution. Sorry, I've missed that there is `#SCROLL` command, because of I referred [this site](https://hitkey.nekokan.dyndns.info/cmdsJP.htm).

It seems that they are [Bemuse's extensions](https://bemuse.ninja/project/docs/bms-extensions/). But I'd like to support them because I'm a contributor of Bemuse. I'll make it an issue.

Anyway, I think this library is still unstable and want to keep `todo!()` until there is no missing major commands unsupported. Currently, a good way to improve I thought is adding debug message about reporting bug to this repository.

MiyakoMeow:
Ah...
I think that there is a situation when the bms file itself has errors in a little part. It's not the fault of this library, but the panic is unacceptable when using this library.
This PR's purpose is to read vaild chart infomation as more as possible.
I'll keep my fork when I want to use this library.
Thanks for replying.

Issue #91, opened 2024-12-28
Base62 Support


MiyakoMeow:
There is an extension: `base62`. `10 + 26 + 26 == 62`

`base62` is appeared in some works in BOF:TT. [For Example](https://manbow.nothing.sh/event/event.cgi?action=More_def&num=286&event=146)

Examples:
- `#WAV0A` 1_1.wav
- `#WAV0a` 1_2.wav

I didn't find the definition page about `base62`.
[An LR2 Base62 Patch Project](https://github.com/SayakaIsBaka/LRbase62)

MikuroXina:
For information: [the specification here](https://docs.google.com/document/u/0/d/e/2PACX-1vTl8zOS3ukl5HpuNsBUlN8rn_ZaNdJSHb8a4se3Z3ap9Y6UJ1nB8LA3HnxWAk9kMTDp0j9orpg43-tl/pub).

PR #99, opened 2025-01-08
fix(bms/lex): use Cow<'a, str>, for performance?

MiyakoMeow:
The improvement I think is almost none.
But it may be a little useful when writing other part.
I create this because of the PR on #91, where changes `Token::Message.message` from `&'a str` to `String`.

### DIFF
--- a/src/lex/token.rs
+++ b/src/lex/token.rs
@@ -398,7 +398,9 @@ impl<'a> Token<'a> {
                 id.make_uppercase();
             }
             Message { message, .. } => {
-                message.make_ascii_uppercase();
+                if message.chars().any(|ch| ch.is_ascii_lowercase()) {
+                    message.to_mut().make_ascii_uppercase();
+                }
             }
             Scroll(id, _) => {
                 id.make_uppercase();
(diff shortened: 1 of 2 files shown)


MikuroXina:
In almost case, this optimization is not helpful for performance, I thought. Most of BMS files are not using base-62 format, then this library allocates memory and make the message uppercase for them.
And sometimes, copying the small string data is cheaper than referencing the large file data.

MikuroXina:
Hmm. It seems that works well but again, the optimization of avoiding clone isn't always improve performance significantly. The reason why I used `&str` in definition of `Token` is that can use the source string with no extra complexity. But allocation for making it uppercase is needed intrinsically at almost cases. So I chose to use `String` instead of `&str` for reducing complexity.
Please understand that I'm not aiming to avoid allocation as possible.

MikuroXina:
Could it be, do you want to reduce allocations when the messages are already uppercase? It can be easily by `Cow` like this:
```rs
if message.chars().any(|ch| ch.is_ascii_lowercase()) {
    message.to_mut().make_ascii_uppercase();
}
```

MikuroXina (approved):

Issue #105, opened 2025-01-10
bms/lex/Cursor: Error when the source str has no \n at the last line.


MiyakoMeow:
- Test code:
```rust
#[test]
fn test_cursor_with_no_ending_return() {
    let text = r"#TITLE Sample";

    let tokens = bms_rs::lex::parse(text).expect("must be parsed");
    let mut tokens_iter = tokens.into_iter();
    assert_eq!(tokens_iter.next().unwrap(), Token::Title("Sample"));
    assert_eq!(tokens_iter.next(), None);
}
```

- Output
```
---- test_cursor_with_no_ending_return stdout ----
thread 'test_cursor_with_no_ending_return' panicked at src\lex\cursor.rs:80:25:
byte index 19 is out of bounds of `#TITLE Sample`
```

PR #106, opened 2025-01-10
Fix Errors about Cursor when parsing source with no "\r\n" in the end.

MiyakoMeow:
#105

### DIFF
--- a/src/lex/cursor.rs
+++ b/src/lex/cursor.rs
@@ -75,7 +75,7 @@ impl<'a> Cursor<'a> {
     pub(crate) fn next_line_remaining(&mut self) -> &'a str {
         let remaining_end = self.source[self.index..]
             .find('\n')
-            .unwrap_or(self.source.len());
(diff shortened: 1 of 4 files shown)


MiyakoMeow:
A bug in Cursor is found and fixed.
The name of methods confuses me before I read the code. I modified them with totally more words and comments.

src/lex/cursor.rs
```diff
@@ -92,7 +98,7 @@ impl<'a> Cursor<'a> {
         self.col
     }
 
-    pub(crate) fn err_expected_token(&self, message: &'static str) -> LexError {
+    pub(crate) fn generate_err_expected_token(&self, message: &'static str) -> LexError {
```
MikuroXina, line 101: Thanks for contribution. It looks almost all good to me. But I'd like to make `generate_err_expected_token` more shorter like `make_err_expected_token`, because it is used frequently.

MikuroXina (approved):

Issue #108, opened 2025-01-10
Full #RANDOM/#SWITCH support


MiyakoMeow:
See [BMS Command memo#ControlFlow](https://hitkey.bms.ms/cmds.htm#CONTROL-FLOW).

PR #109, opened 2025-01-10
add(bms): full random/switch support

MiyakoMeow:
#108


MiyakoMeow:
There is some limitations due to line scan parsing method:
- `#DEF` in front of `#SWITCH` will never activated.
See the test in the end of `nested_switch.rs` for more info.

Sorry for code quality, I just don't know how to write better. The new tests is full of duplicated code.
And the test code in `parse/random.rs` is just debug records.

MiyakoMeow:
Because the usage of let-else statement, the MSRV is now 1.65.

src/parse/random/block.rs
```diff
@@ -0,0 +1,168 @@
+use std::collections::BTreeSet;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(super) enum ControlFlowBlock {
+    Random(RandomBlock),
+    Switch(SwitchBlock),
+}
```
MikuroXina, line 7: It seems that `ControlFlowBlock`'s variants have too responsibilities. I'd like to decompose it for readability. How about like this? I think this helps parsing to be simple.
```rs
pub(super) enum ControlFlowBlock {
    Random {
        // it can be used for warning unreachable branch
        rand_max: u32,
        chosen_value: u32,
    },
    IfBranch {
        // it can be used for warning unreachable `#ELESIF`
        matching_value: u32,
    },
    Switch {
        // it can be used for warning unreachable branch
        rand_max: u32,
        chosen_value: u32,
        // whether skipping tokens until `#ENDSW`
        skipping: bool,
    },
    CaseBranch {
        // it can be used for warning unreachable `#CASE`
        matching_value: u32,
    },
    DefaultBranch,
}
```

MikuroXina (changes requested):

src/lex/token.rs
```diff
@@ -244,7 +253,46 @@ impl<'a> Token<'a> {
...
+                    let case_value = c
+                        .next_token()
+                        .ok_or_else(|| c.make_err_expected_token("switch case value"))?
+                        .parse()
+                        .map_err(|_| c.make_err_expected_token("integer"))?;
+                    Self::Case(case_value)
+                }
+                "#SKIP" => Self::Skip,
+                "#DEFAULT" | "#DEF" => Self::Def,
+                "#ENDSWITCH" | "#ENDSW" => Self::EndSwitch,
```
MikuroXina, line 294: Are `#DEFAULT` and `#ENDSWITCH` used actually? It seems that they are not defined in the specification.

MiyakoMeow: No specification found. I just write a non-shorthand name for taking place. It seems like there is no other command that will take the place.

MikuroXina: To prevent a confusion for the userland, I want to be careful about handling unfamiliar commands. At this stage, I want to put off supporting the non-shorthand names.

MikuroXina (approved):

PR #126, opened 2025-07-16
ControlFlow: Use AST to support more situations, pass the insane test in BMS command memo

MiyakoMeow:


src/bms/parse/random/ast_build.rs
```diff
@@ -0,0 +1,591 @@
...
+                    iter.next();
+                    let (etokens, _has_endif) = parse_if_block_body(iter, error_list);
+                    branches.insert(
+                        0,
+                        IfBranch {
+                            value: 0,
+                            tokens: etokens,
+                        },
+                    );
+                }
```
MikuroXina, line 274: On lexicalize BMS source, the value 0 is allowed for `#IF` and `#ELSEIF`. To prevent occurring ambiguity, value 0 in `#IF` and `#ELSEIF` should be treated as an error.

src/bms/parse/random/ast_build.rs
```diff
@@ -0,0 +1,591 @@
...
+pub(super) struct IfBlock<'a> {
+    pub branches: HashMap<u64, IfBranch<'a>>,
+}
+
+/// The If branch of a If block.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(super) struct IfBranch<'a> {
+    pub value: u64,
+    pub tokens: Vec<Unit<'a>>,
+}
```
MikuroXina, line 50: According to [BMS command memo - `#ELSEIF`](https://hitkey.nekokan.dyndns.info/cmds.htm#ELSEIF), multiple branches for same key can be appeared in a Random block. This data structure can't represent such that enough, because treating them in flat list can't distinguish multiple `#IF`-`#ENDIF` blocks.

src/bms/parse/random/ast_parse.rs
```diff
@@ -0,0 +1,277 @@
...
+                            let mut branch_iter = branch.tokens.clone().into_iter().peekable();
+                            result.extend(parse_control_flow_ast(
+                                &mut branch_iter,
+                                rng,
+                                error_list,
+                            ));
+                            found = true;
+                            break;
+                        }
+                    }
```
MikuroXina, line 53: You can write this more simply:
```suggestion
                    
                    if let Some(else_branch) = if_blocks
                        .iter()
                        .flat_map(|if_block| if_block.branches.get(&0))
                        .next()
                    {
                        let mut branch_iter = else_branch.tokens.clone().into_iter().peekable();
                        result.extend(parse_control_flow_ast(&mut branch_iter, rng, error_list));
                    }
```

MikuroXina (changes requested):

src/bms/parse/random/ast_build.rs
MiyakoMeow: https://hitkey.nekokan.dyndns.info/cmds.htm#ELSEIF

He gives an example, and **red**-notes that, for an `#IF` block, it only choose one branch.
When there is two `#IF` branch with same key in an `#IF` block, the latter would **NEVER** be applied.

PR #127, opened 2025-07-17
Refactor: delete crate::bms pub use

MiyakoMeow:


MiyakoMeow:
#129

MiyakoMeow:
A problem that is not obvious for now, but if several modules added, the pub use will be confusing.
"What the lex and parse module for? BMS? BMSON? or other purpose?"
My opinion is to del this export, at least, set deprecated.
But for now and in the viewable future, this crate has 3 modules at most in root, so... It's OK not modifing it.

PR #128, opened 2025-07-17
add(doc): References of BMS Format

MiyakoMeow:


MikuroXina (changes requested):
Thanks for completing references, I've already forgotten them.

MikuroXina (approved):

Issue #129, opened 2025-07-17
My modification plans


MiyakoMeow:
- [x] Make `Token` includes all the non-empty content of a BMS file. (Add `Token::UnknownCommand("#XXX")` and `Token::NotACommandLine("Not A Command")`) #132 
- [x] Parse token as more as possible, returns a list of warnings. Never return a error. The judge of processing success or not is mainly the user. #132 
  - [x] This crate would add some basic checks, like bpm missing, no notes, no playing notes, `BmsWarning` and `BmsError` type will be added. Same as above, those can be ignored. #134 
- [x] ~~Classify Tokens into ControlFlow, Header, Message and Text.~~(A `Token` is enough, classify will make it complex). 
- [x] ~~Add transition rules for some situations, e.g. `Text("#RONDAM 4")` => `ControlFlow(Random(4))`. This can be added by users, using `Iterator<Item = Fn(Token) => Token>`.~~(User can directly modify `Vec<Token>`)
  - [x] The note channel generation can also be modified by user. A preset config is required? #135 
- [x] Add line sign for each token, for better BMS file debug. #135 Adding just in `LexWarning` is enough. #135
  - [x] Add for Token, to address the line of `ParseWarning` like random block issue. #146 
- [x] BMSON might procress into BMS, when BMS has more functionality. But there will be a problem that track will exceed 999. #145 
- [ ] Add a `chart_process` module, in the same level of `bms` and `bmson` module, to compute the notes position in a view, at a time. And determine which audio to play. This is the most important part for BMS playing. (It will check if the bms is playable.)
- [x] Unifined calling methods for bms parsing. Only use function for one-step parse. #154 
- [x] Can parse AST tree of control flow, so that some advanced bms editors can use this crate. #159

MiyakoMeow:
> Sounds good. But I have some questions:
> 
> * The 4th plan seems impossible, because transforming BMS into Bmson is irreversible.
> * The 5th plan may become out of responsibility of this crate. Whether audio container files can play is important to have knowledge of audio codecs.

The 4th plan I think is possible. Irreversible is not problem.
The 5th plan, I will only do the computing of note positions and fetch the list of the audios to be played at a time. Actual video rendering and audio playing is not in plan.

MikuroXina:
I know. That you said about the 4th is to process a BMS source from a Bmson source, right? Currently, this crate doesn't have features to output BMS source commands, but the parsed data structures derive `Serialize`/`Deserialize`.
[...]

PR #132, opened 2025-07-18
Breaking!: modified the usage of `parse` and `Bms::from_token_stream`, only return warnings

MiyakoMeow:
#129 
1, 2 in partial.

### DIFF
--- a/src/bms/parse.rs
+++ b/src/bms/parse.rs
@@ -37,9 +37,9 @@ pub enum ParseWarning {
     /// The object has required but not defined,
     #[error("undefined object: {0:?}")]
     UndefinedObject(ObjId),
-    /// Parsing is halted because `prompt_handler` returned [`PromptChoice::Halt`].
-    #[error("parsing is halted by prompt handler")]
-    Halted,
+    /// Parsing is warned because `prompt_handler` returned [`DuplicationWorkaround::Warn`].
+    #[error("parsing is warned by prompt handler")]
+    PromptHandlerWarning,
 }
 
 /// type alias of core::result::Result<T, ParseWarning>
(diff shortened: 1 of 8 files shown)


src/bms/lex/token.rs
```diff
@@ -127,7 +129,7 @@ pub enum Token<'a> {
     },
     /// `#MIDIFILE [filename]`. Defines the MIDI file as the BGM. *Deprecated*
     MidiFile(&'a Path),
-    /// Non-empty lines that not starts in # in bms file.
+    /// Non-empty lines that not starts in `['#', '%']` in bms file.
```
MikuroXina, line 133: In fact, the line that starts with `%` is not a command. But `%URL` and `%EMAIL` is currently treated as special tokens reluctantly. We should not recognize other lines such as `%FOO`.

src/bms/parse.rs
```diff
@@ -55,29 +50,46 @@ pub struct Bms {
...
-    pub fn from_token_stream(
-        token_stream: &TokenStream,
+    pub fn from_token_stream<'a>(
+        token_iter: &mut std::iter::Peekable<impl Iterator<Item = &'a Token<'a>>>,
         rng: impl Rng,
         mut prompt_handler: impl PromptHandler,
-    ) -> Result<Self> {
-        let continue_tokens = parse_control_flow(token_stream, rng)?;
+    ) -> BmsParseOutput {
+        let (continue_tokens, mut errors) = parse_control_flow(token_iter, rng);
```
MikuroXina, line 70: It seems that the parameter requires users too many characteristics. How about doing to take only `impl Iterator<Item=&'a Token<'a>>` and create a peekable iterator from it?

MikuroXina (approved):

PR #136, opened 2025-07-19
Bmson: add beatoraja ext

MiyakoMeow:


tests/files_bmson.rs
```diff
@@ -0,0 +1,13 @@
+use bms_rs::bmson::Bmson;
+use std::fs;
+
+#[test]
+fn test_bmson_parse() {
+    let path = "tests/lostokens.bmson";
+    let data = fs::read_to_string(path).expect("failed to read lostokens.bmson");
+    let bmson: Bmson = serde_json::from_str(&data).expect("failed to parse bmson json");
+    // 基本字段断言
```
MikuroXina, line 9: Please write comments in English.

src/bmson/fin_f64.rs
```diff
@@ -25,6 +25,13 @@ impl From<FinF64> for f64 {
     }
 }
 
+impl TryFrom<f64> for FinF64 {
+    type Error = ();
+    fn try_from(value: f64) -> Result<Self, Self::Error> {
+        value.is_finite().then_some(Self(value)).ok_or(())
+    }
+}
```
MikuroXina, line 33: It can provide an Error type like this:
```suggestion
#[derive(Debug, thiserror::Error)]
pub struct TryFromFloatError(pub(crate) ());

impl std::fmt::Display for TryFromFloatError {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        write!(f, "finite number expected")
    }
}

impl TryFrom<f64> for FinF64 {
    type Error = TryFromFloatError;
    fn try_from(value: f64) -> Result<Self, Self::Error> {
        value
            .is_finite()
            .then_some(Self(value))
            .ok_or(TryFromFloatError(()))
    }
}
```

src/bmson/fin_f64.rs
```diff
@@ -35,11 +42,7 @@ impl FinF64 {
     /// Creates a new `FinF64` from `f64` if `float` is finite, otherwise returns `None`.
     #[inline]
     pub fn new(float: f64) -> Option<Self> {
-        if float.is_finite() {
-            Some(Self(float))
-        } else {
-            None
-        }
+        float.is_finite().then_some(Self(float))
```
MikuroXina, line 45: It can be simpler:
```suggestion
        float.try_from().ok()
```

src/bmson.rs
```diff
@@ -129,7 +147,7 @@ pub fn default_mode_hint() -> String {
 
 /// Default relative percentage, 100%.
 pub fn default_percentage() -> FinF64 {
-    FinF64::new(100.0).unwrap()
+    FinF64::new(100.0).expect("Internal error: 100.0 is not a valid FinF64")
```
MikuroXina, line 150: I think that this `expect` with a message doesn't make sense. `100.0` is a finite valid number trivially, so I'd chosen to use `unwrap`.

src/bmson.rs
```diff
@@ -231,6 +273,68 @@ pub struct BgaEvent {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub struct BgaId(pub u32);
 
+/// Beatoraja implementation of long note type.
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, Default,
+)]
```
MikuroXina, line 279:
```suggestion
#[derive(
    Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, Default,
)]
```
`LongNoteType` should not impl `PartialOrd`, because these values don't have meaning about its order.

src/bmson.rs
MiyakoMeow: That is why I signed "Internal error". My habit is never use `unwrap` in codes that should not panic in almost case. If a `unwrap` is required, I would rather use `expected` and sign internal error.

src/bmson/fin_f64.rs
MikuroXina: Sorry, this is typo of `try_into`.

src/bmson.rs
```diff
@@ -378,38 +490,94 @@ impl TryFrom<Bms> for Bmson {
...
+                        });
+                    }
+                    NoteKind::Invisible => {
+                        key_map.entry(note.obj).or_default().push(KeyEvent {
+                            x: note_lane,
+                            y: pulses,
+                        });
+                    }
+                    _ => {
+                        // 普通音符
```
MikuroXina, line 510: Please write comments in English, too.

MikuroXina (changes requested):

MikuroXina (approved):

PR #139, opened 2025-07-25
Bms: use BigUInt & Decimal

MiyakoMeow:


src/bmson/pulse.rs
```diff
@@ -2,47 +2,52 @@
 
 use std::collections::BTreeMap;
 
+use num::traits::SaturatingAdd;
```
MikuroXina, line 5: `u64` has `saturating_add` method (but different signature) since 1.47.0. So I think we don't need to use this.

MikuroXina (changes requested):

src/bms/parse/rng.rs
```diff
@@ -20,3 +33,69 @@ impl<const N: usize> Rng for RngMock<N> {
...
+/// use bms_rs::bms::parse::rng::{Rng, RandRng};
+/// use rand::{rngs::StdRng, SeedableRng};
+/// use num::BigUint;
+///
+/// let mut rng = RandRng(StdRng::seed_from_u64(42));
+/// let n = rng.generate(num::range_inclusive(BigUint::from(1u64), BigUint::from(10u64)));
+/// assert!(n >= BigUint::from(1u64) && n <= BigUint::from(10u64));
+/// ```
+#[cfg(feature = "rand")]
+pub struct RandRng<R: rand::RngCore>(pub R);
```
MikuroXina, line 51:
```suggestion
pub struct RandRng<R>(pub R);
```
In the best practice, you shouldn't add trait bounds for type parameters in a struct definition. It prevents us from adding features that not depends on `RngCore`.

MikuroXina (changes requested):

src/bms/parse/rng.rs
```diff
@@ -1,22 +1,104 @@
...
+    /// - Example:
+    /// ```rust
+    /// use bms_rs::bms::parse::rng::{Rng, RngMock};
+    /// use num::BigUint;
+    ///
+    /// let mut rng = RngMock([BigUint::from(1u64)]);
+    /// let n = rng.generate(num::range_inclusive(BigUint::from(1u64), BigUint::from(10u64)));
+    /// assert!(n >= BigUint::from(1u64) && n <= BigUint::from(10u64));
+    /// ```
+    fn generate(&mut self, range: RangeInclusive<BigUint>) -> BigUint;
```
MikuroXina, line 23: [`RangeInclusive`](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html) I wrote means std's one , that is a superset of num's `RangeInclusive`.

MikuroXina (changes requested):

MikuroXina (changes requested):

MikuroXina (approved):

PR #141, opened 2025-07-28
Bms: add beatoraja & remaining tokens

MiyakoMeow:


MikuroXina (commented):
Minor extensions (such as `#WAVECMD`) by hard-to-find and obsolete BMS players would not be glad to the user to implement features to support. How about to provide them when enabled a feature flag?

src/bms/lex/token.rs
```diff
@@ -381,14 +466,54 @@ impl<'a> Token<'a> {
...
+                        .parse()
+                        .map_err(|_| c.make_err_expected_token("wavcmd value u32"))?;
+                    // Validity check
+                    match param {
+                        WavCmdParam::Pitch if !(0..=127).contains(&value) => {
+                            return Err(c.make_err_expected_token("pitch 0-127"));
+                        }
+                        WavCmdParam::Volume if value > 100 => {
+                            return Err(c.make_err_expected_token("volume 0-100"));
+                        }
```
MikuroXina, line 506: Maybe the volume parameter can be greater than 100. It thought that was discouraged but allowed.

src/bms/lex/token.rs
```diff
@@ -686,24 +898,303 @@ impl<'a> Token<'a> {
...
+                    // Parse xxx.yyy zzzz
+                    let (xy, ms) = if let Some((xy, ms)) = part.split_once(' ') {
+                        (xy, ms)
+                    } else if let Some((xy, ms)) = part.split_once('\t') {
+                        (xy, ms)
+                    } else if let Some((xy, ms)) = part.split_once('\u{3000}') {
+                        (xy, ms)
+                    } else {
+                        return Err(c.make_err_expected_token("stp format xxx.yyy zzzz"));
+                    };
```
MikuroXina, line 952: If you want to extract  spans of `xxx.yyy` and `zzz`, you can use `c.next_token()` instead of using `next_line_remaining` then splitting with whitespaces.

PR #143, opened 2025-07-31
breaking!: Rearrange struct positions

MiyakoMeow:


src/bms/command.rs
```diff
@@ -1,15 +1,15 @@
 //! Definitions of command argument data.
 pub mod channel;
+pub mod graphics;
+pub mod time;
 
-#[cfg(feature = "minor-command")]
+#[allow(unused)]
 use std::time::Duration;
```
MikuroXina, line 7: Please remove the unused import.

src/bms/parse.rs
```diff
@@ -1,27 +1,26 @@
-//! Parser for BMS format. The reason why the implementation separated into lex and parse is the score may contain some randomized elements such as `#RANDOM`. This separation make us able to parse the tokens with the custom random generator cheaply.
+//! BMS: Parse Module.
+//!
+//! Lex => [Parse]
```
MikuroXina, line 3: This doc comment seems to too lack of descriptions. There should be a guide intended for developers to read.

src/bms/parse/prompt.rs
```diff
@@ -4,12 +4,9 @@
...
 use std::path::Path;
 
-use crate::{bms::Decimal, lex::command::ObjId};
+use crate::bms::{Decimal, command::*};
 
-use super::{
-    ParseWarning, Result,
-    header::{AtBgaDef, BgaDef, Bmp, ExRankDef, ExWavDef},
-};
+use super::{ParseWarning, Result, model::def::*};
```
MikuroXina, line 9: Please don't use wildcard import in the file using cfg feature attributes. It may be introduced namespace collision, especially switched feature flags.

src/bms/parse/random.rs
```diff
@@ -1,13 +1,16 @@
...
 
 use thiserror::Error;
 
-use super::{ParseWarning, rng::Rng};
+use super::ParseWarning;
 use crate::bms::{lex::token::Token, parse::BmsParseTokenIter};
 
-use self::ast_build::*;
-use self::ast_parse::*;
+pub use self::rng::*;
```
MikuroXina, line 12: Please don't re-export wildcard imports. It makes that this module strongly depends on `rng` module's items. I want to avoid that.

src/bms/parse/random.rs
```diff
@@ -1,13 +1,16 @@
+//! Random Part.
```
MikuroXina, line 1: This doc comment has no informations, and this module doesn't have guide you have to write. So you can remove this.

MikuroXina (changes requested):

src/bms/parse/random.rs
MiyakoMeow: Removing this will cause document warnings, but I don't know what to write.

MiyakoMeow:
The `Bms` model struct is re-arranged.

src/bms/parse/model.rs
```diff
@@ -0,0 +1,1227 @@
...
+                    self.bpm_changes.insert(bpm_change.time, existing);
+                }
+                DuplicationWorkaround::UseNewer => {
+                    // Already inserted the newer one
+                }
+                DuplicationWorkaround::Warn => {
+                    return Err(ParseWarning::PromptHandlerWarning);
+                }
+            }
+        }
```
MikuroXina, line 987: This may occur two insertions. How about checking whether it was contained with BTreeMap's `contains_key` method?

src/bms/parse/random.rs
MikuroXina: I think it's ok to suppress the warning here.

MikuroXina (changes requested):

src/bms/parse/model/obj.rs
```diff
@@ -0,0 +1,261 @@
...
+impl Ord for ScrollingFactorObj {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.time.cmp(&other.time)
+    }
+}
+
+/// An object to change spacing factor between notes with linear interpolation.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct SpeedFactorObj {
```
MikuroXina, line 204: As the description, this object represents the change of spacing factor, not speed (against the command name). So I want to name it `SpacingFactorObj`.

src/bms/command.rs
```diff
@@ -1,15 +1,18 @@
...
-pub use channel::Channel;
-
-#[cfg(feature = "minor-command")]
-use crate::time::ObjTime;
-
-use super::{Result, cursor::Cursor};
+/// Export defs
+pub use channel::*;
+pub use graphics::*;
+pub use time::*;
```
MikuroXina, line 15: It may be introduced hard complexity because of mixing make submodules public and re-exporting their items. I like to choose either one.

MikuroXina (changes requested):

src/bms/parse/model/obj.rs
MiyakoMeow: However, the original command is `#SPEED`. I would like to keep the original "speed" naming. The description is from the  old code and I forgot to modify it.

src/bms/command.rs
MiyakoMeow: I will delete these re-exports. And, how about adding an `prelude` module for users, like other crates?

src/bms/parse/model/obj.rs
MikuroXina: But it is not "speed factor" just `#SPEED`. If you want to name it after the command, you should choose `SpeedObj`.

src/bms/parse/model/obj.rs
```diff
@@ -198,31 +205,31 @@ impl Ord for ScrollingFactorObj {
     }
 }
 
-/// An object to change spacing factor between notes with linear interpolation.
+/// An object to change speed factor of notes (a.k.a. spacing factor between notes) with linear interpolation.
```
MikuroXina, line 208:
```suggestion
/// An object to change spacing factor among notes with linear interpolation.
```
According to [https://bemuse.ninja/project/docs/bms-extensions/](Bemuse's extension doc), this object changes the notes spacing, not scrolling speed. So I'd like to avoid "speed factor" to describe the meaning of this object.

MikuroXina (changes requested):

src/bms/parse/model/obj.rs
```diff
@@ -0,0 +1,268 @@
...
+use crate::bms::{
+    Decimal,
+    command::{
+        ObjId,
+        channel::{Channel, Key, NoteKind, PlayerSide},
+        time::{ObjTime, Track},
+    },
+};
+
+/// An object on the score.
```
MikuroXina, line 11:
```suggestion
/// An object playing sound on the score.
```
We might as well improve that this doc to be more specific.

MikuroXina (changes requested):

MikuroXina (approved):

PR #145, opened 2025-08-03
feat: impl From<Bmson> for Bms

MiyakoMeow:


src/bmson.rs
```diff
@@ -672,3 +675,245 @@ impl TryFrom<Bms> for Bmson {
         })
     }
 }
+
+impl From<Bmson> for Bms {
+    fn from(value: Bmson) -> Self {
```
MikuroXina, line 680: This process isn't considered about it is lossy (loses precise data) conversion for Bmson. Even thought parsing BMS from the source text produces warning, but this function ignores all conflicts of insertions.
I think it should be changed into a method like `from_bmson_lossy` that returns warnings too.

MikuroXina:
The root cause of consideration is `bmson::Bmson` permits constructing its data without checking whether it conforms to the specification. So then the conversion method may receive the invalid Bmson data.

src/bmson/bmson_to_bms.rs
```diff
@@ -0,0 +1,393 @@
...
+                .scroll_defs
+                .insert(scroll_def_id, factor.clone());
+
+            bms.arrangers
+                .scrolling_factor_changes
+                .insert(time, ScrollingFactorObj { time, factor });
+        }
+
+        // Convert sound channels to notes
+        let mut obj_id_counter = 1u16;
```
MikuroXina, line 150: Manually counting and checking id may mistake How about add this:
```
#[derive(Debug)]
struct ObjIdIssuer(u16);

impl ObjIdIssuer {
    const fn new() -> Self {
        Self(1)
    }
}

impl Iterator for ObjIdIssuer {
    type Item = ObjId;
    fn next(&mut self) -> Option<Self::Item> {
        const MAX_ID: u16 = 62 * 62;
        if self.0 > MAX_ID {
            return None;
        }
        let id = self.0;
        self.0 += 1;
        create_obj_id_from_u16(id).ok()
    }
}
```

MikuroXina (changes requested):

src/bmson/bms_to_bmson.rs
```diff
@@ -0,0 +1,356 @@
...
+                Some(BpmEvent {
+                    y: converter.get_pulses_at(bpm_change.time),
+                    bpm: FinF64::new(bpm_change.bpm.clone().to_f64().unwrap_or(120.0))
+                        .unwrap_or_else(|| {
+                            warnings.push(BmsToBmsonWarning::InvalidBpm);
+                            FinF64::new(120.0).expect("Internal error: 120.0 is not a valid FinF64")
+                        }),
+                })
+            })
+            .collect();
```
MikuroXina, line 105: This `filter_map` can be replaced with `map` like this:
```suggestion
        let bpm_events =
            self.arrangers
                .bpm_changes
                .values()
                .map(|bpm_change| BpmEvent {
                    y: converter.get_pulses_at(bpm_change.time),
                    bpm: FinF64::new(bpm_change.bpm.clone().to_f64().unwrap_or(120.0))
                        .unwrap_or_else(|| {
                            warnings.push(BmsToBmsonWarning::InvalidBpm);
                            FinF64::new(120.0).expect("Internal error: 120.0 is not a valid FinF64")
                        }),
                })
                .collect();
```

src/bmson/bms_to_bmson.rs
```diff
@@ -0,0 +1,356 @@
...
+                            | Key::Key13
+                            | Key::Key14
+                            | Key::ScratchExtra
+                            | Key::FootPedal => 0,
+                        } + match note.side {
+                            PlayerSide::Player1 => 0,
+                            PlayerSide::Player2 => 8,
+                        },
+                    )
+                    .and_then(|lane| NonZeroU8::new(lane));
```
MikuroXina, line 220:
```suggestion
                    .and_then(NonZeroU8::new);
```
This redundant closure can be reduced.

MikuroXina (changes requested):
From clippy warnings:

PR #146, opened 2025-08-03
feat: Add position info for `Token` & `ParseWarning`

MiyakoMeow:


src/bms/lex.rs
```diff
@@ -82,16 +85,22 @@ pub fn parse_lex_tokens_with_channel_parser<'a>(
     let mut tokens = vec![];
     let mut warnings = vec![];
     while !cursor.is_end() {
-        match Token::parse(&mut cursor, channel_parser) {
-            Ok(token) => tokens.push(token),
+        match TokenContent::parse(&mut cursor, channel_parser) {
+            Ok(content) => tokens.push(Token {
+                content,
+                row: 0,
+                col: 0,
```
MikuroXina, line 92:
```suggestion
                row: cursor.line(),
                col: cursor.col(),
```
Could you retrieve these `row` and `col` info from `cursor`? `0:0` should be discouraged.

src/bms/parse.rs
```diff
@@ -43,8 +40,36 @@ pub enum ParseWarning {
...
+            self.content, self.row, self.col
+        )
+    }
+}
+
+impl std::error::Error for ParseWarning {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&self.content)
+    }
+}
```
MikuroXina, line 72:
```suggestion
#[derive(Debug, Clone, PartialEq, Eq, Hash, Error)]
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
pub struct ParseWarning {
    /// The content of the parse warning.
    #[source]
    pub content: ParseWarningContent,
    /// The row (line number) where the warning occurred.
    pub row: usize,
    /// The column (character position) where the warning occurred.
    pub col: usize,
}

impl std::fmt::Display for ParseWarning {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(
            f,
            "{} at line {}, column {}",
            self.content, self.row, self.col
        )
    }
}
```
You can write this more simply with `thiserror::Error` derive macro.

src/bms/parse/random/ast_build.rs
```diff
@@ -106,18 +107,19 @@ pub(super) fn build_control_flow_ast<'a>(
...
             let (unit, errs) = parse_switch_block(iter);
             Some((unit, errs))
         }
-        Token::Random(_) | Token::SetRandom(_) => {
+        Random(_) | SetRandom(_) => {
             let (unit, errs) = parse_random_block(iter);
             Some((unit, errs))
         }
-        token if !token.is_control_flow_token() => {
+        _ if !token.content.is_control_flow_token() => {
```
MikuroXina, line 122:
```suggestion
        content if !content.is_control_flow_token() => {
```
It can shorten with the discarded one.

MikuroXina (changes requested):

PR #147, opened 2025-08-03
doc(README): Re-add usage

MiyakoMeow:


README.md
```diff
@@ -61,6 +58,71 @@ This will be placed as:
    |-----------------|
 ```
 
+## Usage
+
+- **NOTE**: BMS files now is almost with `SHIFT-JIS` encoding. It's recommended to use [`encoding_rs`](https://crates.io/crates/encoding_rs) crate to parse raw file to `Cow<str>`, which is a compatible type of `&str`, using `AsRef::as_ref`.
```
MikuroXina, line 63:
```suggestion
- **NOTE**: BMS files now is almost with Shift_JIS encoding. It's recommended to use [`encoding_rs`](https://crates.io/crates/encoding_rs) crate to parse raw file to `Cow<str>`, which is a compatible type of `&str`, using `AsRef::as_ref`.
```
The strict encoding name is here.

MikuroXina (changes requested):

README.md
MikuroXina: Yes, could you fix that too?

MikuroXina (approved):

PR #149, opened 2025-08-05
Bms: last add tokens

MiyakoMeow:


MikuroXina (changes requested):
Almost of enum defined should not implement `PartialOrd`/`Ord` because they don't have its order conceptually. Also after they implemented `PartialOrd`, reordering variants will be a breaking change.

src/bms/command/time.rs
```diff
@@ -19,16 +21,32 @@ pub struct ObjTime {
...
+            track += numerator / denominator;
+            numerator %= denominator;
+        }
+        // If numerator is 0, set numerator to 0 and denominator to 1, and return.
+        if numerator == 0 {
+            return Self {
+                track: Track(track),
+                numerator: 0,
+                denominator: 1,
+            };
```
MikuroXina, line 44:
```suggestion
```
`0.gcd(&denominator)` will be `0`, so you can remove this `if` branch about the numerator iss zero.

src/bms/command/time.rs
```diff
@@ -19,16 +21,32 @@ pub struct ObjTime {
...
-            eprintln!("warning: track 000 detected");
+    pub fn new(mut track: u64, mut numerator: u64, mut denominator: u64) -> Self {
+        // If denominator is 0, set numerator to 0 and denominator to 1, and return.
+        if denominator == 0 {
+            return Self {
+                track: Track(track),
+                numerator: 0,
+                denominator: 1,
+            };
+        }
```
MikuroXina, line 32: The reason why it panics when `denominator == 0` is setting `denominator` zero is to prevent from mistaking about programming. In most cases, `ObjTime::new` is called with track index, index in the id sequence, and length of the id sequence. So I thought it should be discourageed; to pass `0` as `denominator` lacks check about the application logic.

MikuroXina (changes requested):

src/bms/command/time.rs
MiyakoMeow: That is the problem. if the `gcd` value is `0`, that will occur "divide 0".

src/bms/command/minor_command.rs
```diff
@@ -151,6 +151,56 @@ pub struct SwBgaEvent {
...
+/// BM98 #ExtChr extended character customization event.
+///
+/// Used for #ExtChr command, implements custom UI element image replacement.
+/// - sprite_num: character index to replace [0-1023]
+/// - bmp_num: BMP index (hex to decimal, or -1/-257, etc.)
+/// - start_x/start_y: crop start point
+/// - end_x/end_y: crop end point
+/// - offset_x/offset_y: offset (optional)
+/// - abs_x/abs_y: absolute coordinate (optional)
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
MikuroXina, line 163:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
`ExtChrEvent` should not derive `PartialOrd`/`Ord` because the fields about coordinate don't have its order.

MikuroXina (changes requested):

MikuroXina (approved):

PR #150, opened 2025-08-08
feat!: Move out AST part

MiyakoMeow:


MiyakoMeow:
Sorry for large PR.
Originally I just want to move out the ast part, and move the value checking part to ast_parse, but the duplication of "line, col" is crazy, so that I created the `PositionWrapper` wrapper.
Moving out AST part is for some bms tools to modify the control flow of BMS. After this PR, I will develop `AST -> Token` and `Token -> String`.
It may be better if reading commits in order.
The Modifications is below:
- Move out AST Parts
- Use `PositionWrapper` to replace `row, col` definition.
- Rename back some types that have named before, like `Token`, `ParseWarning` as enum.
- Unifined BMS/BMSON parse interface. Now the parse steps are all in `crate::bms`.
- Rewrite channel converting in `crate::bms::command::channel`.

MikuroXina:
Ugh, I have so many design considerations about `PositionWrappper` and parse interface. So could you split this PR into them:

- Add `PositionWrapper` (I thought it's better to name it `SourcePosMixin`).
  - Should it implement `Deref`/`DerefMut`? Its public fields `row` and `column` may conflict to the contained.
  - …
- Unify parse interface.
  - Is `parse_bmson` unified to BMS's one? It seems returning a `serde_json`'s error.
  - …
- Tidying up and apply a new `PositionWrapper` in terms' definition such as AST, token and others.

PR #151, opened 2025-08-12
feat!: Impl text pos mixin

MiyakoMeow:


src/bms/lex/token.rs
MikuroXina, line 1: It doesn't seem good idea that make users to append `SourcePosMixin<..>` for each type. It may be annoying and make too many changes. I think it's better to wrap like:
```rs
pub type Token<'a> = SourcePosMixin<TokenContent<'a>>;
```

MikuroXina (changes requested):

MiyakoMeow:
How about this?
```
pub enum Token
pub type TokenWithPos = SourcePosMixin<Token>
```

PR #153, opened 2025-08-14
feat: new channel convert method

MiyakoMeow:
I would add an easier way for note channel converting later.

### DIFF
--- a/src/bms/command/channel.rs
+++ b/src/bms/command/channel.rs
@@ -183,109 +183,142 @@ pub fn read_channel_beat(channel: &str) -> Option<Channel> {
     Some(Channel::Note { kind, side, key })
 }
 
-/// Reads a key from a character. (For PMS)
-fn get_key_pms_bme_type(key: char) -> Option<Key> {
-    use Key::*;
-    Some(match key {
-        '1' => Key1,
-        '2' => Key2,
-        '3' => Key3,
-        '4' => Key4,
-        '5' => Key5,
-        '6' => Key8,
-        '7' => Key9,
-        '8' => Key6,
-        '9' => Key7,
-        _ => return None,
-    })
+/// Key channel family for different key mapping schemes.
+///
+/// Mappings used by each family:
+/// - Beat:
+///   - Chars: '1'..'7','6' scratch, '7' free zone, '8'->Key6, '9'->Key7
+///   - Char -> (PlayerSide, Key):
+///     - '1'..'5' => (Player1, Key1..Key5), '6' => (Player1, Scratch),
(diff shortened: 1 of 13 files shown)


src/bms/command/channel.rs
```diff
@@ -183,109 +178,142 @@ pub fn read_channel_beat(channel: &str) -> Option<Channel> {
...
+///   - Beat -> this: (P2,Key2..Key5) remapped to (P1,Key6..Key9); (P1,Key1..Key5) unchanged
+///   - This -> Beat: Key6..Key9 => (P2,Key2..Key5); Key1..Key5 => (P1,Key1..Key5)
+/// - BeatNanasi:
+///   - Beat -> this: FreeZone=>FootPedal
+///   - This -> Beat: FootPedal=>FreeZone
+/// - DscOctFp:
+///   - Beat -> this: (P2,Key1)=>FootPedal, (P2,Key2..Key7)=>Key8..Key13, (P2,Scratch)=>ScratchExtra; (P1,Key1..Key7|Scratch) unchanged; side becomes P1
+///   - This -> Beat: reverse of above
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ModeKeyChannel {
```
MikuroXina, line 202: I think that it's not good about making key mapping as an enum. Because this makes hard to add, modify and extend the mapping.
For maintainability, they should be each own type and implement the common trait like `KeyMapping` that converts the side and key (or from the channel number).

MikuroXina (changes requested):

src/bms/command/channel.rs
```diff
@@ -183,107 +178,194 @@ pub fn read_channel_beat(channel: &str) -> Option<Channel> {
...
+    ///
+    /// This method takes a (PlayerSide, Key) pair in this mode's format and converts
+    /// it to the equivalent (PlayerSide, Key) pair in Beat mode format.
+    fn to_beat(&mut self, side: PlayerSide, key: Key) -> (PlayerSide, Key);
+
+    /// Convert from Beat mode format to this mode's format.
+    ///
+    /// This method takes a (PlayerSide, Key) pair in Beat mode format and converts
+    /// it to the equivalent (PlayerSide, Key) pair in this mode's format.
+    fn map_from_beat(&mut self, side: PlayerSide, key: Key) -> (PlayerSide, Key);
```
MikuroXina, line 197: To avoid confusion of tuple of side and key, I'd like to treat the Beat mode format as own intermediate representation type like `pub struct BeatModeMap(pub PlayerSide, pub Key);`. Then these methods should handle it.

MikuroXina (changes requested):

PR #154, opened 2025-08-14
refactor!: use struct-binding function for single step

MiyakoMeow:


src/bms/lex.rs
```diff
@@ -57,44 +57,75 @@ pub(crate) type Result<T> = core::result::Result<T, LexWarning>;
...
 }
 
-/// Analyzes and converts the BMS format text into [`TokenStream`].
-pub fn parse_lex_tokens<'a>(source: &'a str) -> BmsLexOutput<'a> {
-    parse_lex_tokens_with_channel_parser(source, &read_channel_beat)
+/// A list of tokens.
+/// This is a wrapper of [`Vec<TokenWithPos<'a>>`] that provides some additional methods.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+pub struct TokenStream<'a> {
```
MikuroXina, line 69: It seems that this struct provide no feature to us. I thought it's adding more complexity rather than improving. How about just unifying `parse_lex_tokens` and `parse_lex_tokens_with_channel_parser`?

MikuroXina (changes requested):
Commit 594f9ac changes behavior of the parser and unrelated, so it shouldn't be included in a refactor PR.

src/bms/lex.rs
MiyakoMeow: Let's see `Bms` type. About usage change this, `TokenStream` is a tips for the step's inner returning type.
I think `BmsLexOutput { tokens: TokenStream, .. } = TokenStream::parse_lex(..)` is more unified with `BmsParseOutput { bms: Bms, .. } = Bms::from_token_stream(..)`.

MikuroXina (approved):

PR #157, opened 2025-08-14
fix: Delete "extend message" impl

MiyakoMeow:


MikuroXina:
Notice: The prefix of commit message is set for [that release-please allows recognize Conventional Commit messages](https://github.com/googleapis/release-please?tab=readme-ov-file#how-should-i-write-my-commits). So if you expected to express changes needed to add to the release note, you should choose appropriate prefix such as `feat:` or `fix:` (however, I may make changes when merging depending on the situation).
If you're not sure to select which one, please don't write it down on the PR title. I'll pick them up.

MikuroXina:
`#EXT` extension is came from [Bemuse's extension](https://bemuse.ninja/project/docs/bms-extensions/#extension-lines), and I don't know what the meaning is. The reason why you removed it is lack of the specification of that extension, right?

MikuroXina (approved):

PR #159, opened 2025-08-16
feat: Move out ast part

MiyakoMeow:


MiyakoMeow:
Maybe the workflow file need to update?

src/lib.rs
```diff
@@ -122,4 +122,4 @@
 pub mod bms;
 pub mod bmson;
 
-pub use bms::{command, lex, parse};
+pub use bms::{ast, command, lex, parse, parse_bms, prelude};
```
MikuroXina, line 125: Why did you re-export `parse_bms` function here? I think it should be in `prelude` rather than the crate root.

MikuroXina (changes requested):

PR #160, opened 2025-08-16
doc: Split for `README.md` and `lib.rs`

MiyakoMeow:


src/lib.rs
```diff
@@ -60,60 +62,6 @@
...
-//! 003|--|--------------|
-//!    |  |03            |
-//!    |  |03            |
-//!    |  |              |
-//!    |  |03            |
-//! 002|--|03------------|
-//!    |  |  []  []  []  |
-//!    |()|[]  []  []  []|
-//!    |-----------------|
-//! ```
```
MikuroXina, line 116: I think the crate root doc comments should be a single truth source of this crate's information. I wrote "About the format" section for the user to focus reading in README, but the README isn't (and shouldn't be) very important as a doc of BMS format. Beacuse of this circumstance, please don't remove this section.

src/lib.rs
```diff
@@ -2,6 +2,8 @@
 //!
 //! Be-Music Source, called BMS for short, is a file format devised by Urao Yane in 1998 for a simulator of the game Beatmania by KONAMI. This describes what and when notes are arranged and its music metadata. It is a plain text file with some "command" lines starting with `#` character.
 //!
+//! For detailed information about the BMS format and this crate, see `README.md` in [github:MikuroXina/bms-rs](https://github.com/MikuroXina/bms-rs).
+//!
```
MikuroXina, line 6:
```suggestion
```
It's not recommended to add a link to GitHub's content. Because it will be connected to the latest one regardless what the crate's version is. And referencing another doc from the truth source should be discouraged.

src/lib.rs
```diff
@@ -83,11 +83,11 @@
 //! #00211:0303030303
 //! ```
 //!
-//! ### Header command
+//! ## Header command
 //!
 //! Header commands are used to express the metadata of the music or the definition for note arrangement.
 //!
-//! ### Message command
+//! ## Message command
```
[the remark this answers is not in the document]
MiyakoMeow: Checked again, it's better not change it.

PR #168, opened 2025-08-19
fix: split `PlayerSide` for `ids_by_key`

MiyakoMeow:


src/bms.rs
```diff
@@ -103,7 +109,11 @@ pub fn parse_bms(source: &str) -> BmsOutput {
     let mut warnings: Vec<BmsWarning> = lex_warnings.into_iter().map(BmsWarning::Lex).collect();
 
     // Parse BMS using default RNG and prompt handler
+    #[cfg(feature = "rand")]
     let rng = RandRng(StdRng::from_os_rng());
+    #[cfg(not(feature = "rand"))]
+    let rng = crate::ast::rng::RngMock([BigUint::one()]);
+
```
MikuroXina, line 116: It's not recommended to change behavior of the function by feature flags. If you want to add feature to change randomness, you can add an optional parameter like:
```rs
pub fn parse_bms(source: &str, rng: Option<impl Rng>) -> BmsOutput {
```

src/bms/parse/model.rs
```diff
@@ -200,12 +201,12 @@ pub struct Notes {
...
-    /// All note objects, indexed by ObjId. #XXXYY:ZZ... (note placement)
-    pub objs: HashMap<ObjId, Vec<Obj>>,
+    pub bgms: MultiMap<ObjTime, ObjId>,
+    /// All note objects, indexed by ObjId (like BMSON's `sound_channel`). #XXXYY:ZZ... (note placement)
+    pub objs: HashMap<ObjId, MultiMap<ObjTime, Obj>>,
     /// Index for fast key lookup. Used for LN/landmine logic.
-    /// Maps each key (lane) to a sorted map of times and object IDs for efficient note lookup.
-    pub ids_by_key: HashMap<Key, BTreeMap<ObjTime, ObjId>>,
+    /// Maps each ([`PlayerSide`], [`Key`]) pair to a sorted map of times and [`ObjId`]s for efficient note lookup.
+    pub ids_by_key: HashMap<(PlayerSide, Key), BTreeMap<ObjTime, ObjId>>,
```
MikuroXina, line 209: As logical concept, `Key` represents the logical lane of notes. So using `Key` along `PlayerSide` will break responsibility of `Key`, I thought.

src/bms/parse/model.rs
```diff
@@ -200,12 +201,12 @@ pub struct Notes {
     pub wav_files: HashMap<ObjId, PathBuf>,
     // objects stored in obj is sorted, so it can be searched by bisection method
     /// BGM objects, indexed by time. #XXX01:ZZ... (BGM placement)
-    pub bgms: BTreeMap<ObjTime, Vec<ObjId>>,
-    /// All note objects, indexed by ObjId. #XXXYY:ZZ... (note placement)
-    pub objs: HashMap<ObjId, Vec<Obj>>,
+    pub bgms: MultiMap<ObjTime, ObjId>,
```
MikuroXina, line 204: `MultiMap<K, V>` uses `HashMap<K, Vec<V>>` internally, so I think it will get performance worse on scanning bgm notes chronologically.

src/bms/parse/model.rs
```diff
@@ -200,12 +201,12 @@ pub struct Notes {
...
+    /// All note objects, indexed by ObjId (like BMSON's `sound_channel`). #XXXYY:ZZ... (note placement)
+    pub objs: HashMap<ObjId, MultiMap<ObjTime, Obj>>,
```
MikuroXina, line 206: `objs` is master data source of all note objects. To seek notes efficiently, I added `ids_by_key` abd `bgms`. Please use `MultiMap` for `objs` like this:
```rs
pub objs: MultiMap<ObjId, Obj>,
```

src/bms/parse/model.rs
MiyakoMeow: If not modified, it will influence DP mode's revelant logic. It is truly a bug.

MikuroXina (approved):

PR #169, opened 2025-08-20
feat: Impl bms to token

MiyakoMeow:


src/bms/parse/to_tokens.rs
MikuroXina, line 1: I think this feature should not be in `parse` module. The responsibility of `parse` modules is to convert BMS source into the music notes. I prefer to place the feature that creates another BMS source from music notes into another module declared from the crate root.

src/bms/unparse.rs
```diff
@@ -0,0 +1,738 @@
...
+        // Convert scroll factor to hex format
+        let factor_u8 = factor.to_u8().unwrap_or(1);
+        format!("{:02X}", factor_u8)
+    }
+
+    fn speed_to_message(&self, factor: Decimal) -> String {
+        // Convert speed factor to hex format
+        let factor_u8 = factor.to_u8().unwrap_or(1);
+        format!("{:02X}", factor_u8)
+    }
```
MikuroXina, line 608: `#SCROLL` and `#SPEED` allow fractional value, so you should cast decimal into `f64` instead of `u8`.

MikuroXina (changes requested):

MiyakoMeow:
Sorry for unchecked. Now there is an complete impl.

src/bms/unparse.rs
```diff
@@ -0,0 +1,906 @@
...
+            }
+
+            // Convert base BPM
+            if let Some(base_bpm) = &arrangers.base_bpm {
+                tokens.push(Token::BaseBpm(base_bpm.clone()));
+            }
+        }
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
MikuroXina, line 402:
```suggestion
```
Now you can remove suppressing this warning.

MikuroXina (changes requested):

PR #170, opened 2025-08-20
feat: Impl ast to token

MiyakoMeow:


MikuroXina (changes requested):

src/lib.rs
```diff
@@ -54,6 +54,25 @@
...
+//!
+//! let source = std::fs::read_to_string("tests/files/lilith_mx.bms").unwrap();
+//! let LexOutput { tokens, lex_warnings } = TokenStream::parse_lex(&source);
+//! assert_eq!(lex_warnings, vec![]);
+//!
+//! // Build AST from tokens
+//! let AstBuildOutput { root, ast_build_warnings } = AstRoot::from_token_stream(&tokens);
+//! assert_eq!(ast_build_warnings, vec![]);
+//!
+//! // Extract tokens back from AST (inverse operation)
```
MikuroXina, line 72:
```suggestion
//! // Extract tokens back from AST
```
Hmm, strictly speaking, it is not *inverse* operation. I think it's better to remove it.

MikuroXina (changes requested):

MikuroXina (approved):

PR #171, opened 2025-08-20
feat: impl parse_bmson

MiyakoMeow:


MiyakoMeow:
Is the `zlib` library's license having problems with Apache-2.0?

src/bmson/parse.rs
```diff
@@ -0,0 +1,903 @@
...
+        genre: "Unknown".to_string(),
+        mode_hint: "beat-7k".to_string(),
+        chart_name: String::new(),
+        level: 1,
+        init_bpm: FinF64::new(120.0).unwrap_or_else(|| {
+            warnings.push(BmsonWarning::InvalidFieldValue(
+                "Invalid init_bpm, using default 120.0".to_string(),
+            ));
+            FinF64::new(120.0).unwrap()
+        }),
```
MikuroXina, line 892:
```suggestion
        init_bpm: FinF64::new(120.0).unwrap(),
```
It seems that lacks self-consistency. You are using `unwrap` for `FinF64::new(120.0)`, but it is a handling error of `FinF64::new(120.0)`. So if this code works, the warning will not occur.

MikuroXina (changes requested):

PR #176, opened 2025-08-25
add: AstParseWarning

MiyakoMeow:


src/bms/ast/ast_extract.rs
```diff
@@ -187,7 +187,8 @@ mod tests {
         let random_block = Unit::RandomBlock {
             value: BlockValue::Set {
                 value: BigUint::from(1u32),
-            },
+            }
+            .into_wrapper_manual(0, 0),
```
MikuroXina, line 191: You should use `into_wrapper` to copy the source position of `value`.
```suggestion
            .into_wrapper(&value),
```

MikuroXina (changes requested):

Issue #181, opened 2025-08-25
Ready for fancy source warnings/errors


MikuroXina:
Most of error types in this crate has source position information. This can be applied to some fancy diagnostics library such as [ariadne](https://crates.io/crates/ariadne), but some error types still lack information about span of the source.

Supporting features about fancy diagnostics, let's improve the incomplete error types.

PR #182, opened 2025-08-25
feat: make ast structure public, make the AstParseWarning get the pos

MiyakoMeow:


src/bms/ast/ast_extract.rs
```diff
@@ -28,18 +28,19 @@ pub(super) fn extract_units<'a>(
...
-        BlockValue::Random { max } => max.clone(),
-        BlockValue::Set { value } => value.clone(),
+    let (row, col) = (value.row(), value.column());
+    let random_value = match value.into_content() {
+        BlockValue::Random { max } => max,
+        BlockValue::Set { value } => value,
     };
 
-    let random_token = SourcePosMixin::new(Token::Random(random_value), 0, 0);
+    let random_token = Token::Random(random_value).into_wrapper_manual(row, col);
```
MikuroXina, line 43: You can use `intro_wrapper` to avoid decomposing `value`'s `row` and `col` like this:
```rs
    let random_value = match value.content().clone() {
        BlockValue::Random { max } => max,
        BlockValue::Set { value } => value,
    };

    let random_token = Token::Random(random_value).into_wrapper(&value);
```
Other lines are same.

MikuroXina (changes requested):

src/bms/ast/ast_extract.rs
MiyakoMeow: #179

MikuroXina: Yes, it has an undocumented semantic of the source position of block value. It may be interpreted as the source of range of generating random values. But literally, it can be written with `into_wrapper` as the same proccess.

MikuroXina (approved):
> But literally, it can be written with `into_wrapper` as the same proccess.

That I said is just an improvement, not a fix of #179. And changes you made is too. Also `extract_units` has deteriorated about its return value. So the issue can't be resolved in this PR (but mergable). Did you intend that?

MiyakoMeow:
Now #179 should be consider as solved by this PR.

src/bms/ast/structure.rs
```diff
@@ -0,0 +1,75 @@
...
+    },
+    /// A Switch block.
+    /// Like C++ Programming Language, Switch block can contain multiple Case branches, and a Def branch.
+    /// If there is no other Case branch activated, Def branch will be activated.
+    /// When executing, the tokens, from the activated branch, to Skip/EndSwitch, will be executed.
+    SwitchBlock {
+        /// The value of the Switch block.
+        value: SourcePosMixin<BlockValue>,
+        /// The Case branches in the Switch block.
+        cases: Vec<CaseBranch<'a>>,
```
MikuroXina, line 29: To preserve source position information, you may need these fields:
```suggestion
        /// The ENDIF position in the BMS source if exists.
        end_if: Option<SourcePosMixin<()>>,
    },
    /// A Switch block.
    /// Like C++ Programming Language, Switch block can contain multiple Case branches, and a Def branch.
    /// If there is no other Case branch activated, Def branch will be activated.
    /// When executing, the tokens, from the activated branch, to Skip/EndSwitch, will be executed.
    SwitchBlock {
        /// The value of the Switch block.
        value: SourcePosMixin<BlockValue>,
        /// The Case branches in the Switch block.
        cases: Vec<CaseBranch<'a>>,
        /// The ENDSW position in the BMS source.
        end_sw: SourcePosMixin<()>,
```

MikuroXina (changes requested):
The reason why I avoid using `into_wrapper_manual` is for #181. It needs that `SourcePosMixin` have to become holding index span of a BMS source string, so it will be hard to refactor codes operating row and column manually.

src/bms/ast/ast_extract.rs
```diff
@@ -12,13 +12,21 @@ pub(super) fn extract_units<'a>(
     for unit in units {
         match unit {
             Unit::TokenWithPos(token) => {
-                tokens.push(token.clone());
+                tokens.push((*token).clone());
```
MikuroXina, line 15:
```suggestion
                tokens.push(token.clone());
```
This change is needless due to auto-deref.

src/bms/ast/structure.rs
```diff
@@ -0,0 +1,81 @@
...
+    },
+}
+
+/// The If block of a Random block. Should contain If/EndIf, can contain ElseIf/Else.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct IfBlock<'a> {
+    /// The branches of the If block.
+    pub branches: BTreeMap<BigUint, SourcePosMixin<Vec<Unit<'a>>>>,
+    /// The ENDIF position in the BMS source if exists for this IfBlock.
+    pub end_if: Option<SourcePosMixin<()>>,
```
MikuroXina, line 60:
```suggestion
    /// The ENDIF position in the BMS source if exists for this IfBlock.
    pub end_if: SourcePosMixin<()>,
```
For technical reasons, I think it should not be an `Option`. It's better that `end_if` expresses the source position of the end of the if block. Now, it's ok to fallback to position of the last token when there is no matching `#ENDIF`.

MikuroXina (changes requested):

Issue #173, opened 2025-08-24
Split validity check from parsing BMS source


MikuroXina:
While editing `Bms` struct's data, it can be modified note objects regardless of its validity. It's good to edit and save the BMS file, but bother to check whether it can be played correctly.

Then, I want to split the current validity check from the parsing process and make it be able to invoke as `Bms`'s method.

PR #183, opened 2025-08-25
feat: Impl validity for bms

MiyakoMeow:
#173

### DIFF
--- a/src/bms.rs
+++ b/src/bms.rs
@@ -35,7 +35,6 @@ use self::{
         ParseOutput, ParseWarningWithPos,
         check_playing::{PlayingCheckOutput, PlayingError, PlayingWarning},
         model::Bms,
-        validity::{ValidityCheckOutput, ValidityError, ValidityWarning},
     },
 };
 
@@ -65,12 +64,6 @@ pub enum BmsWarning {
     /// An error for playing.
     #[error("Error: playing: {0}")]
     PlayingError(#[from] PlayingError),
-    /// A warning for validity.
-    #[error("Warn: validity: {0}")]
-    ValidityWarning(#[from] ValidityWarning),
-    /// An error for validity.
-    #[error("Error: validity: {0}")]
-    ValidityError(#[from] ValidityError),
 }
 
 /// Output of parsing a BMS file.
@@ -149,21 +142,6 @@ pub fn parse_bms_with_rng(source: &str, rng: impl Rng) -> BmsOutput {
 
     // Convert playing errors to BmsWarning
     warnings.extend(playing_errors.into_iter().map(BmsWarning::PlayingError));
-
-    let ValidityCheckOutput {
-        validity_warnings,
-        validity_errors,
-    } = bms.check_validity();
-
-    // Convert validity warnings to BmsWarning
-    warnings.extend(
-        validity_warnings
-            .into_iter()
-            .map(BmsWarning::ValidityWarning),
-    );
-
-    // Convert validity errors to BmsWarning
-    warnings.extend(validity_errors.into_iter().map(BmsWarning::ValidityError));
 
     BmsOutput { bms, warnings }
 }
--- a/src/bms.rs
+++ b/src/bms.rs
@@ -35,6 +35,7 @@ use self::{
         ParseOutput, ParseWarningWithPos,
         check_playing::{PlayingCheckOutput, PlayingError, PlayingWarning},
         model::Bms,
+        validity::{ValidityCheckOutput, ValidityError, ValidityWarning},
     },
 };
 
@@ -64,6 +65,12 @@ pub enum BmsWarning {
     /// An error for playing.
     #[error("Error: playing: {0}")]
     PlayingError(#[from] PlayingError),
+    /// A warning for validity.
+    #[error("Warn: validity: {0}")]
+    ValidityWarning(#[from] ValidityWarning),
+    /// An error for validity.
+    #[error("Error: validity: {0}")]
+    ValidityError(#[from] ValidityError),
 }
 
 /// Output of parsing a BMS file.
@@ -142,6 +149,21 @@ pub fn parse_bms_with_rng(source: &str, rng: impl Rng) -> BmsOutput {
 
     // Convert playing errors to BmsWarning
     warnings.extend(playing_errors.into_iter().map(BmsWarning::PlayingError));
+
+    let ValidityCheckOutput {
+        validity_warnings,
+        validity_errors,
+    } = bms.check_validity();
+
+    // Convert validity warnings to BmsWarning
+    warnings.extend(
+        validity_warnings
+            .into_iter()
+            .map(BmsWarning::ValidityWarning),
+    );
+
+    // Convert validity errors to BmsWarning
+    warnings.extend(validity_errors.into_iter().map(BmsWarning::ValidityError));
 
     BmsOutput { bms, warnings }
 }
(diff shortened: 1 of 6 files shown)


src/bms.rs
MikuroXina, line 1: I didn't intent to add validity check on parsing. I'd like this to be optional. So you don't need to add features to this file.

MikuroXina (changes requested):

src/bms/parse/validity.rs
```diff
@@ -0,0 +1,614 @@
...
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum ValidityEmpty {
+    /// Text content is empty at the time.
+    #[error("Empty text content at {0:?}")]
+    EmptyTextAt(ObjTime),
+    /// Option content is empty at the time.
+    #[cfg(feature = "minor-command")]
+    #[error("Empty option content at {0:?}")]
+    EmptyOptionAt(ObjTime),
+}
```
MikuroXina, line 29: Also I want to exclude `ValidityEmpty` because this doesn't influence to correct playback. To normalize them is the userland editors'.

src/bms/parse/validity.rs
```diff
@@ -0,0 +1,614 @@
...
+pub enum ValidityMissing {
+    /// A note references an ObjId without a corresponding WAV definition.
+    #[error("Missing WAV definition for note object id: {0:?}")]
+    MissingWavForNote(ObjId),
+    /// A BGM references an ObjId without a corresponding WAV definition.
+    #[error("Missing WAV definition for BGM object id: {0:?}")]
+    MissingWavForBgm(ObjId),
+    /// A BGA change references an ObjId without a corresponding BMP/EXBMP definition.
+    #[error("Missing BMP definition for BGA object id: {0:?}")]
+    MissingBmpForBga(ObjId),
```
MikuroXina, line 44: Please remove the prefix named its enum type like this:
```suggestion
    #[error("Missing WAV definition for note object id: {0:?}")]
    WavForNote(ObjId),
    /// A BGM references an ObjId without a corresponding WAV definition.
    #[error("Missing WAV definition for BGM object id: {0:?}")]
    WavForBgm(ObjId),
    /// A BGA change references an ObjId without a corresponding BMP/EXBMP definition.
    #[error("Missing BMP definition for BGA object id: {0:?}")]
    BmpForBga(ObjId),
```

src/bms/parse/validity.rs
```diff
@@ -0,0 +1,614 @@
...
+    #[cfg(feature = "minor-command")]
+    #[error("Empty option content at {0:?}")]
+    EmptyOptionAt(ObjTime),
+}
+
+/// Missing-related validity entries.
+#[non_exhaustive]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Error)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum ValidityMissing {
```
MikuroXina, line 35: There are other missing cases to consider (even only major commands) more:

- Missing BPM change `#BPMxx`
- Missing scroll stop `#STOPxx`

src/bms/parse/validity.rs
```diff
@@ -0,0 +1,614 @@
...
+        /// Player side where the mapping is placed.
+        side: PlayerSide,
+        /// Key lane where the mapping is placed.
+        key: Key,
+        /// Timestamp of the mapping.
+        time: ObjTime,
+        /// Mapped object id.
+        mapped: ObjId,
+    },
+}
```
MikuroXina, line 80: This check comes from lack of encapsulation of `Bms`, so it should be removed. We need to improve `Bms`'s APIs not to happen such situations.

MikuroXina (changes requested):

src/bms/parse/validity.rs
```diff
@@ -0,0 +1,465 @@
...
+            if !self.scope_defines.stop_defs.contains_key(id) {
+                missing.push(ValidityMissing::StopDef(*id));
+            }
+        }
+
+        // 3) Placement/overlap checks for notes on lanes.
+        //      - Visible notes in section 000
+        //      - Overlap: visible single vs single (same time, same lane)
+        //      - Overlap: visible single within long interval (same lane)
+        //      - Overlap: landmine within long interval -> warn once at long start
```
MikuroXina, line 156: Oops, there should be an overlapping check about landmine vs single.

src/bms/parse/validity.rs
```diff
@@ -0,0 +1,465 @@
...
+            let mut lane_objs = objs;
+            lane_objs.sort_unstable_by_key(|o| o.offset);
+
+            // Build LN intervals by pairing consecutive Long notes
+            let mut long_times: Vec<ObjTime> = lane_objs
+                .iter()
+                .filter(|o| o.kind == NoteKind::Long)
+                .map(|o| o.offset)
+                .collect();
+            long_times.sort_unstable();
```
MikuroXina, line 189:
```suggestion
```
It is needless to sort because `lane_objs` has already sorted and its order has unchanged.

src/bms/parse/validity.rs
```diff
@@ -0,0 +1,465 @@
...
+            }
+
+            // Helper: check if a time is within [s, e]
+            let time_overlaps_any_ln = |t: ObjTime| -> Option<(ObjTime, ObjTime)> {
+                for (s, e) in &ln_intervals {
+                    if t >= *s && t <= *e {
+                        return Some((*s, *e));
+                    }
+                }
+                None
```
MikuroXina, line 231: You can improve the performance like this algorithm with `long_times`:
1. Search the first item index `end_i` which is greater than or equal to `t` in `long_times` (or return `None` if not found).
    - Since `long_times` is sorted, it can be searched with [`slice::partition_point`](https://doc.rust-lang.org/std/primitive.slice.html#method.partition_point) rapidly like `long_times.partition_point(|&end| end < t)`.
2. Check whether `end_i` is an odd number. If it is not, there is no sections contains `t` so return `None`.
3. Return the section `(long_times[end_i - 1], long_times[end_i])`.

MikuroXina (changes requested):

src/bms/parse/validity.rs
```diff
@@ -0,0 +1,529 @@
...
+                }
+                if has_landmine {
+                    invalid.push(ValidityInvalid::LandmineOverlapsLongAtStart {
+                        side,
+                        key,
+                        ln_start: *s,
+                        ln_end: *e,
+                    });
+                }
+            }
```
MikuroXina, line 300: It can be done same as "Overlap single vs long". Then `ln_intervals` will become needless.
```suggestion
            for o in &lane_objs {
                if o.kind == NoteKind::Landmine
                    && let Some((s, e)) = time_overlaps_any_ln(o.offset)
                    && s == o.offset
                {
                    invalid.push(ValidityInvalid::LandmineOverlapsLongAtStart {
                        side,
                        key,
                        ln_start: s,
                        ln_end: e,
                    });
                }
            }
```

MikuroXina (changes requested):

src/bms/parse/validity.rs
```diff
@@ -0,0 +1,604 @@
...
+        /// Start time of the long note interval.
+        ln_start: ObjTime,
+        /// End time of the long note interval.
+        ln_end: ObjTime,
+    },
+    /// A landmine note overlaps a long note interval; warn only at the long start point.
+    #[error(
+        "Landmine overlaps a long note starting at {ln_start:?} (side={side:?}, key={key:?}; ln_end={ln_end:?})"
+    )]
+    LandmineOverlapsLongAtStart {
```
MikuroXina, line 86:
```suggestion
    OverlapLandmineWithLongAtStart {
```
I'd like to fit the naming as other variants.

MikuroXina (changes requested):

PR #185, opened 2025-08-28
feat!: NoteChannel

MiyakoMeow:


src/bms/command/channel.rs
```diff
@@ -314,58 +435,10 @@ fn read_channel_general(channel: &str) -> Option<Channel> {
...
-    /// Create a new [`KeyMapping`] from a [`PlayerSide`] and [`Key`].
-    fn new(side: PlayerSide, key: Key) -> Self;
-    /// Get the PlayerSide from this KeyMapping.
-    fn side(&self) -> PlayerSide;
-    /// Get the [`Key`] from this [`KeyMapping`].
-    fn key(&self) -> Key;
-    /// Deconstruct into a [`PlayerSide`], [`Key`] tuple.
-    fn into_tuple(self) -> (PlayerSide, Key);
-}
+// read_channel_beat has been removed, unified to use `read_channel` (returns NoteChannel) and `read_channel_general` (general enum only).
```
MikuroXina, line 444: Could you remove these removal comments? They are traceable by git history.

src/bms.rs
```diff
@@ -99,19 +100,19 @@ pub struct BmsOutput {
...
-    parse_bms_with_rng(source, rng)
+    parse_bms_with_rng::<BeatKey>(source, rng)
 }
 
 /// Parse a BMS file from source text using a custom random number generator.
 ///
 /// This function provides a convenient way to parse a BMS file in one step.
 /// It uses the default channel parser and a custom random number generator.
-pub fn parse_bms_with_rng(source: &str, rng: impl Rng) -> BmsOutput {
+pub fn parse_bms_with_rng<T: PhysicalKey + Default>(source: &str, rng: impl Rng) -> BmsOutput<T> {
```
MikuroXina, line 115: The cause of required `Default` is `Bms` and its children are auto-deriving `Default`. `#[derive(..)]` requires the same trait to its type parameters, so `T` is required `Default`. However, data of `Bms` and its children don't seem required `T: Default`. So you can impl them manually to remove the parameter trait bound `Default`.

MikuroXina (changes requested):

PR #187, opened 2025-08-30
Impl fancy source warnings errors

MiyakoMeow:
#181


MikuroXina (changes requested):
No, please modify `SourcePosMixin` into hold index span `(start, end)` to represent corresponding range of the source string. If you did it, diagnostics will be very easy to implement.

src/bms/diagnostics.rs
MikuroXina, line 1: You don't need to provide features that calculates row and column of the source. It will be done by ariadne. I think that we only required to provide the error message and index span.

src/bms/lex/cursor.rs
```diff
@@ -156,34 +153,38 @@ fn test1() {
...
+
+    // Test end of input
+    assert_eq!(cursor.next_token(), None);
 }
 
 #[test]
 fn test2() {
     const SOURCE: &str = r"
-        #TITLE 花たちに希望を [SP ANOTHER]
+        #TITLE Hope to Flowers [SP ANOTHER]
```
MikuroXina, line 179: Why did you weaken this test case? The new index-based system must work on non-ASCII characters too.

MikuroXina (changes requested):
(DEVELOPMENT NOTICE: It seems that this your branch based on so old commit, but mergeable. If you want to tidy up the situation, you can use `git rebase`.)

MikuroXina (changes requested):

MikuroXina (changes requested):

src/bms/lex/token.rs
```diff
@@ -104,7 +104,7 @@ pub enum Token<'a> {
     Charset(&'a str),
     /// `#COMMENT [string]`. Defines the text which is shown in the music select view. This may or may not be surrounded by double-quotes.
     Comment(&'a str),
-    /// `#DEF`. Starts a case scope if any `#CASE` had not matched to the generated random number. It must be placed in the end of the switch scope. See also [`TokenWithPos::Switch`].
+    /// `#DEF`. Starts a case scope if any `#CASE` had not matched to the generated random number. It must be placed in the end of the switch scope. See also [`TokenWithRange::Switch`].
```
MikuroXina, line 107: We might as well to fix the links?
```suggestion
    /// `#DEF`. Starts a case scope if any `#CASE` had not matched to the generated random number. It must be placed in the end of the switch scope. See also [`Token::Switch`].
```

src/bms/lex/token.rs
```diff
@@ -159,7 +159,7 @@ pub enum Token<'a> {
     },
     /// `#GENRE [string]`. Defines the genre of the music.
     Genre(&'a str),
-    /// `#IF [u32]`. Starts an if scope when the integer equals to the generated random number. This must be placed in a random scope. See also [`TokenWithPos::Random`].
+    /// `#IF [u32]`. Starts an if scope when the integer equals to the generated random number. This must be placed in a random scope. See also [`TokenWithRange::Random`].
```
MikuroXina, line 162:
```suggestion
    /// `#IF [u32]`. Starts an if scope when the integer equals to the generated random number. This must be placed in a random scope. See also [`Token::Random`].
```

src/bms/lex/cursor.rs
```diff
@@ -47,12 +48,23 @@ impl<'a> Cursor<'a> {
...
+    /// Peek the next token with range.
+    #[allow(unused)]
+    pub(crate) fn peek_next_token_with_range(&self) -> Option<(std::ops::Range<usize>, &'a str)> {
+        let ret = self.peek_next_token_range();
+        if ret.is_empty() {
+            return None;
+        }
+        Some((ret.clone(), &self.source[ret]))
+    }
+
```
MikuroXina, line 60:
```suggestion
```
I think it is ok to remove the unused.

MikuroXina (changes requested):

src/bms/lex/token.rs
```diff
@@ -1189,7 +1189,10 @@ impl<'a> Token<'a> {
             command if command.starts_with('#') => Self::UnknownCommand(c.next_line_entire()),
             _not_command => Self::NotACommand(c.next_line_entire()),
         };
-        Ok(token)
+
+        // Calculate the full range of this token (from command start to current cursor position)
+        let token_range = command_range.start..c.index();
+        Ok(SourceRangeMixin::new(token, token_range))
```
MikuroXina, line 1195:
```suggestion
        Ok(SourceRangeMixin::new(token, command_range))
```
I think the command's range should not include trailing whitespaces. So you don't need to extend the range.

MikuroXina (changes requested):

src/bms/lex/token.rs
MikuroXina: Oh, I got it. Its range is what consisting of all the command.

PR #188, opened 2025-08-30
feat: impl channel key bindings with Bms

MiyakoMeow:


src/bms/command/channel.rs
```diff
@@ -175,6 +173,87 @@ pub enum PlayerSide {
...
+        if value.len() != 2 {
+            return Err(value);
+        }
+        let mut chars = value.bytes();
+        let [Some(ch1), Some(ch2), None] = [chars.next(), chars.next(), chars.next()] else {
+            return Err(value);
+        };
+        Self::try_from([ch1, ch2]).map_err(|_| value)
+    }
+}
```
MikuroXina, line 225: Instead of impl `TryFrom<&str>`, you should impl [`FromStr` trait](https://doc.rust-lang.org/std/str/trait.FromStr.html). It has a better type inference.

src/bms/command/channel.rs
```diff
@@ -175,6 +173,87 @@ pub enum PlayerSide {
...
+        Into::<u16>::into(value) as u64
+    }
+}
+
+impl ChannelId {
+    /// Instances a special null id, which means the rest object.
+    pub const fn null() -> Self {
+        Self([0, 0])
+    }
+
```
MikuroXina, line 250: According to [<CHANNEL> in BMS command memo](https://hitkey.nekokan.dyndns.info/cmds.htm#CHANNEL), the channel `00` is a retired number. It doesn't mean rest or such special things.

MikuroXina (changes requested):

src/bms/command/channel.rs
```diff
@@ -175,6 +174,85 @@ pub enum PlayerSide {
...
+                s.len()
+            ));
+        }
+        let mut chars = s.bytes();
+        let [Some(ch1), Some(ch2), None] = [chars.next(), chars.next(), chars.next()] else {
+            return Err(format!("Failed to parse channel characters from '{}'", s));
+        };
+        Self::try_from([ch1, ch2]).map_err(|_| format!("Invalid channel characters: '{}'", s))
+    }
+}
```
MikuroXina, line 229: Please avoid to specify error message strings as the error type. Instead, you should add an error type like this:
```rs
#[derive(Debug, Clone, PartialEq, Eq, Hash, Error)]
pub enum ChannelIdParseError {
    #[error("channel id must be exactly 2 ascii characters, got `{0}`")]
    ExpectedTwoAsciiChars(String),
    #[error("channel id must be an alpha numeric to parse as base 62, got {0}")]
    InvalidAsBase62(String),
}
```

MikuroXina (changes requested):

PR #201, opened 2025-09-11
add: parse_bmson, serde_path_to_error version

MiyakoMeow:


src/bmson.rs
```diff
@@ -316,9 +323,58 @@ pub struct KeyEvent {
...
+pub fn parse_bmson(json: &str) -> Result<Bmson<'_>, serde_path_to_error::Error<serde_json::Error>> {
+    // First try to parse with serde_path_to_error for better error reporting
+    let jd = &mut serde_json::Deserializer::from_str(json);
+    let result: Result<Bmson, _> = serde_path_to_error::deserialize(jd);
+
+    result
+}
+
+#[cfg(feature = "diagnostics")]
+impl ToAriadne for serde_path_to_error::Error<serde_json::Error> {
```
MikuroXina, line 352: I think we should not impl `ToAriadne` for `serde_path_to_error::Error`. Because it is a structural error information. Users will be confused with the meaningless span and line diagnosis.

MikuroXina (changes requested):

src/bmson.rs
```diff
@@ -138,8 +145,8 @@ pub struct BmsonInfo {
 
 /// Default mode hint, beatmania 7 keys.
 #[must_use]
-pub fn default_mode_hint() -> String {
-    "beat-7k".into()
+pub fn default_mode_hint() -> Cow<'static, str> {
+    Cow::Borrowed("beat-7k")
```
MikuroXina, line 149:
```suggestion
pub fn default_mode_hint() -> &'static str {
    "beat-7k"
```
I think libraries should provide the most widest type for extendability and should not be weaken as a `Cow`.

MikuroXina (changes requested):

MikuroXina:
> I did the rebase. And now the only question is, use `String` or `Cow` in `Bmson`?

I respect your intention that using `Cow`, and I suggested a change to improve working with `default_mode_hint` (that will be good whether we chose `Cow` or not). But if I wrote `Bmson`, I will choose the `String` way.

MiyakoMeow:
So, if the function return type concerns, I'll modify `Bmson` back to use `String`.

MikuroXina (approved):

PR #202, opened 2025-09-11
feat!(converter): deal with iterator of `Key` or `(PlayerSide, Key)`

MiyakoMeow:


src/bms/ast/rng.rs
MikuroXina, line 1: Why did you deprecate `JavaRandom`? It is an option of `Rng` to emulate beatoraja's behavior, isn't it?

src/bms/command/channel/converter.rs
```diff
@@ -1,118 +1,74 @@
...
+/// - Difference from [`super::mapper::KeyLayoutMapper`]:
+///   - [`super::mapper::KeyLayoutMapper`] can convert between different key channel modes. It's two-way.
+///   - [`KeyMappingConverter`] can convert into another key layout. It's one-way.
+///   - [`KeyMappingConverter`] operates on iterators of `(ObjTime, KeyMapping)` pairs, preserving timing information.
+pub trait KeyMappingConverter {
+    /// Convert an iterator of `(ObjTime, KeyMapping)` pairs to another key layout.
+    fn convert<T: KeyMapping>(
+        &mut self,
+        mappings: impl Iterator<Item = (ObjTime, T)>,
+    ) -> impl Iterator<Item = (ObjTime, T)>;
```
MikuroXina, line 20: Why do `convert` require `ObjTime` as a first item of the tuple while preserving it? It should be removed like:
```suggestion
///   - [`KeyMappingConverter`] operates on iterators of `(ObjTime, KeyMapping)` pairs, preserving timing information.
pub trait KeyMappingConverter {
    /// Convert an iterator of `KeyMapping` to another key layout's one.
    fn convert<T: KeyMapping>(
        &mut self,
        mappings: impl Iterator<Item = T>,
    ) -> impl Iterator<Item = T>;
```

MikuroXina (changes requested):

src/bms/command/channel/converter.rs
MikuroXina: From a musical point of view, the overlapped notes should be kept to play the music correctly. This would be warned by our diagnostics, but handling them (such as merging into one) are delegated to userland.
Anyway, the features to shuffle notes won't (and shouldn't) change timing of notes, I expect.

MiyakoMeow: OK. Is it suitable for adding several traits for degrees of ability? But it's a challenge for naming them.

MikuroXina (approved):

PR #203, opened 2025-09-17
mv: diagnostics and impl

MiyakoMeow:


src/diagnostics.rs
MikuroXina, line 1: Just make to sure; why did you move `diagnostics` into the crate root? It is a feature for BMS source file structure now, isn't it? Are you planning to reuse `SimpleSource` for diagnostics of Bmson?

MiyakoMeow: Oh, I have forgot. If the `chumsky` version of `parse_bmson` applied, the movement of `diagnostics` makes sense.
But for now, it's truly only used in `bms`. I'll do the movement

Issue #11, opened 2023-04-15
Support bmson format


MikuroXina:
[Bmson format](https://bmson-spec.readthedocs.io/en/master/doc/index.html) is a file format based on JSON and used on modern BMS players. It is helpful for us to support to serialize/deserialize the format.

To implement this, using [`serde_json`](https://github.com/serde-rs/json) will be good.

PR #207, opened 2025-09-21
fix!: ObjTime should use NonZeroU64

MiyakoMeow:


src/bms/parse.rs
```diff
@@ -508,6 +508,9 @@ impl<T: KeyLayoutMapper> Bms<T> {
                     if bpm == 0 {
                         continue;
                     }
+                    let denominator = NonZeroU64::new(denominator).ok_or_else(|| {
+                        ParseWarning::SyntaxError("denominator cannot be zero".to_string())
+                    })?;
```
MikuroXina, line 513: Maybe this is not a syntax error but an internal error. If it occurred, the message was empty since building `Token::Message` went wrong.

MikuroXina (changes requested):

MiyakoMeow:
Fixes #208

src/bms/parse.rs
MiyakoMeow: No. If it happens, it seems like a line of simply `#00011:`.

MiyakoMeow: I think it's unneed to add extra checks, because when meets illegal char in message, there will be another ParseWarning.
Or, do you want to parse this situation without the non-ASCII char influence? 
`#00111:一0000`
This will need more effort which is not related to this PR.

src/bmson.rs
```diff
@@ -158,8 +164,39 @@ pub fn default_percentage() -> FinF64 {
...
+                ))
+            })
+        }
+        Some(positive @ 1..) => NonZeroU64::new(positive as u64).ok_or_else(|| {
+            serde::de::Error::custom(format!(
+                "Cannot convert resolution {} to valid NonZeroU64",
+                positive
+            ))
+        }),
+    }
```
MikuroXina, line 199: It may be difficult to support #208 right now. According to [Numbers section in JSON specification](https://datatracker.ietf.org/doc/html/rfc8259#section-6), "This specification allows implementations to set limits on the range and precision of numbers accepted". So we may be needed to take care about such limits like `[-(2**53)+1, (2**53)-1]`.
Errors and/or warnings about this deserialization is going complex, but it's stray from this PR which aims to change `ObjTime`. I want to resolve the issue in another one.

MiyakoMeow: Reverted the commit.

PR #209, opened 2025-09-22
feat: bmson resolution deserialize

### LINKED ISSUE #208: Non-positive resolution in Bmson
According to [Bmson specification](https://bmson-spec.readthedocs.io/en/master/doc/index.html):

> `resolution :: unsigned long`
>
> This is the number of pulses per one quarter note in a 4/4 measure.
>
> Default value is `240`.
>
> - `resolution` must be > 0.
>   - If 0, `null` or `undefined`, use the default value.
>   - If negative, take the absolute value.

So we must accept non-positive value (also `null` and `undefined`) and fallback to the default value.

MiyakoMeow:
Fixes #208


src/bmson.rs
```diff
@@ -382,35 +380,41 @@ where
...
+                rounded
+            );
+            if rounded == 0.0 {
+                return Ok(default_resolution_nonzero());
+            }
+
+            // Ensure the rounded value fits in u64 exactly
+            if rounded > (u64::MAX as f64) {
+                return Err(E::custom(format!("Resolution value too large: {}", v)));
+            }
```
MikuroXina, line 414: According to [WebIDL definition of Bmson](https://bmson-spec.readthedocs.io/en/master/doc/index.html#format-overview) and other descriptions, `resolution` is defined as an `unsigned long`. So I think non-integer (non-zero `av.fract()`) should be an error. Or does some Bmson player round them?

MikuroXina (changes requested):

PR #212, opened 2025-09-23
feat: Impl bms to token

MiyakoMeow:


src/bms/model/notes.rs
```diff
@@ -120,6 +120,14 @@ impl<T> Notes<T> {
             .map(|(idx, obj)| (WavObjArenaIndex(idx), obj))
     }
 
+    /// Returns the iterator having all of the notes in the original insertion order.
+    ///
+    /// This reflects the order notes were pushed into the arena during parsing, which
+    /// corresponds to the lexical order of `Token::Message` entries in the source.
+    pub fn all_notes_insertion_order(&self) -> impl Iterator<Item = &WavObj> {
+        self.arena.0.iter()
+    }
```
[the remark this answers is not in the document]
MiyakoMeow: Wait, When I write visible notes that points to a non-exist `wav_id`, you hope it to disappear?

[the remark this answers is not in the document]
MikuroXina: Hmm, it should be discouraged as a playable BMS (because most of BMS events disqualifies it if it has too many unassigned notes), but it's just a valid BMS.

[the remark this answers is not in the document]
MikuroXina: Yes, the danling note ([`WavObj::dangling`](https://github.com/MikuroXina/bms-rs/blob/3b0ee8144c8e60cbc68160390c4c548651a154c2/src/bms/model/obj.rs#L46)) is an unassigned BGM note. So it doesn't affect to anything musical and playing.

src/bms/unparse.rs
```diff
@@ -16,8 +126,49 @@ impl<T: KeyLayoutMapper> Bms<T> {
     #[must_use]
     pub fn unparse<'a>(&'a self) -> Vec<Token<'a>> {
         let mut tokens: Vec<Token<'a>> = Vec::new();
+        let mut needs_base62_token = false;
```
MikuroXina, line 129: To check whether base 62 is needed, how about to ask `ObjIdManager`s whether the ids in base 62 range are used? Checking with this boolean flag seems dangerous and annoying.

src/bms/command.rs
```diff
@@ -217,6 +206,81 @@ impl ObjId {
...
+}
+
+/// Enum representing the source of a message value that can be converted to [char; 2]
+#[derive(Clone, Copy, Debug)]
+pub enum MessageValue {
+    /// Value derived from an ObjId
+    ObjId(ObjId),
+    /// Value derived from a u8
+    U8(u8),
+}
```
MikuroXina, line 265: Value of message command is not exhausted with this. For example, BM98 supports decimal fraction number for channel `02` (section length) like this:
```text
#00102:12.375f
```
Also, there may be more formats on message commands in future. So I think we cannot use this approach which always converting into `char`s.

MikuroXina (changes requested):

MiyakoMeow:
Finished

src/bms/command.rs
MikuroXina: If you want to add a this type (message value that can be converted to `[char; 2]`), it lacks the responsibility too much, I thought. The ways to convert into a `String` should be implemented by each ones separately.

MikuroXina (changes requested):

src/bms/command.rs
```diff
@@ -294,3 +330,192 @@ impl TryFrom<u8> for LnMode {
...
+    #[must_use]
+    pub fn get_unused_ids(&self) -> &VecDeque<ObjId> {
+        &self.unused_ids
+    }
+
+    /// Extract the ObjIdManager (legacy method for compatibility)
+    #[must_use]
+    pub fn extract(self) -> (HashMap<&'a K, ObjId>, HashSet<ObjId>, VecDeque<ObjId>) {
+        (self.value_to_id, self.used_ids, self.unused_ids)
+    }
```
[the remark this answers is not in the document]
MikuroXina: I want `ObjIdManager` to do not control `create_token`. Could you use `get_or_new_id` instead of `get_or_allocate_id`?

src/bms/command.rs
```diff
@@ -294,3 +328,151 @@ impl TryFrom<u8> for LnMode {
...
+        if let Some(&id) = self.value_to_id.get(key) {
+            id
+        } else if let Some(new_id) = self.unused_ids.pop_front() {
+            self.used_ids.insert(new_id);
+            self.value_to_id.insert(key, new_id);
+            new_id
+        } else {
+            ObjId::null()
+        }
+    }
```
MikuroXina, line 406: I'd like `get_or_new_id` to return the `Option<ObjId>` to handle it by the callee when failed the allocate.
```suggestion
    pub fn get_or_new_id(&mut self, key: &'a K) -> Option<ObjId> {
        if let Some(&id) = self.value_to_id.get(key) {
            Some(id)
        } else if let Some(new_id) = self.unused_ids.pop_front() {
            self.used_ids.insert(new_id);
            self.value_to_id.insert(key, new_id);
            Some(new_id)
        } else {
            None
        }
    }
```

MikuroXina (changes requested):

src/bms/unparse.rs
```diff
@@ -0,0 +1,1313 @@
...
+                        ObjIdAllocResult::Assigned(id) => Some(id),
+                        ObjIdAllocResult::New(id) => {
+                            let def_token = token_creator(id, key);
+                            late_def_tokens.push(def_token);
+                            Some(id)
+                        }
+                        ObjIdAllocResult::Full => None,
+                    }
+                })
+                .flatten();
```
MikuroXina, line 1070: It can be written as below if `get_or_new_id` returns `Option<ObjId>`:
```suggestion
            let id = id_allocation
                .as_mut()
                .and_then(|(token_creator, key_extractor, manager)| {
                    let key = key_extractor(event);
                    if manager.is_assigned(key) {
                        manager.get_or_new_id(key)
                    } else if let Some(new) = manager.get_or_new_id(key) {
                        late_def_tokens.push(token_creator(new, key));
                        Some(new)
                    } else {
                        None
                    }
                });
```

MikuroXina: I don't want to return whether the id has just allocated now, because it leaks the `ObjIdManager`'s responsibility and internal state.

MikuroXina (changes requested):

PR #213, opened 2025-09-27
feat: impl base36

MiyakoMeow:


MikuroXina (changes requested):
Acccording to [the Base 62 specification doc that we are referencing](https://github.com/MikuroXina/bms-rs/blob/3b0ee8144c8e60cbc68160390c4c548651a154c2/README.md?plain=1#L83), there is no command `#BASE 36`. We must interpret the BMS source as Base 36 if and only if `#BASE 62` is not written.

MikuroXina:
> Maybe #BASE 36 is for LR2, nanasi or older player?

Older players don't know the `#BASE` extension, so they will ignore the command and interpret BMS source in case-insensitive. It doesn't matter for them if there is no `#BASE` command.

> I just want to avoid this warning.

I think that we must keep it to be warned. I think `#BASE 62` has the meaning if as is, and the number other than 62 is unexpected.

The command is made for marking object ids in the BMS source as case-sensitive. Otherwise we must parse object ids in case-insensitive. Therefore, the invalid commands such as `#BASE 36` misunderstood the purpose of the extension and should be discouraged.

MikuroXina:
I found [the code which produces `#BASE 36` on BMSE's source](https://kuroki-narumi.github.io/BmsMaker3-Manual/bms_defSort.html). Also there is [an editor which supports another `#BASE` and can output `#BASE 16`/`#BASE 36`/`#BASE 62`](https://kuroki-narumi.github.io/BmsMaker3-Manual/bms_defSort.html). (The potential of `#BASE 16` is [said on the FAQ](https://kuroki-narumi.github.io/BmsMaker3-Manual/faq.html#%E5%AE%9A%E7%BE%A9%E6%95%B0%E3%81%AF%E8%B6%B3%E3%82%8A%E3%82%8B%E3%81%AF%E3%81%9A%E3%81%AA%E3%81%AE%E3%81%ABZZ%E5%AE%9A%E7%BE%A9%E3%81%AB%E5%A4%89%E6%8F%9B%E3%81%95%E3%82%8C%E3%81%AA%E3%81%84) too.)

Ok, we might be needed to extend `#BASE` command. Could you add `BaseType::Base16` variant too?

MiyakoMeow:
How to impl Base16? is it required to convert `ObjId` variants into `00..=FF` range?

MikuroXina:
Uh, I was re-designing the id system and found that allowing `#BASE 16` can be problematic. It's ok to not to do now.

MikuroXina (approved):

MikuroXina:
For you information; [the parser library used on beatoraja makes warning if `#BASE` has a number other than 62](https://github.com/exch-bms2/jbms-parser/blob/ca7b666c7446d2775fc539f652b493f778de50a5/src/bms/model/BMSDecoder.java#L595).

MiyakoMeow:
Wait, the message didn't handle right

MikuroXina (changes requested):
Sorry for changing again and again but importantly, object ids doesn't need to interpret as an integer. Its string and hash are only important to match (with equality of case-sensitive or case-insensitive). So we don't need to support features using base have what the number.

At this point, there is no reason to supress warning for `#BASE 16` and `#BASE 36`, I thought. Also there is no document to justify the extension of `#BASE` having number other than 62.

What I am considering is that we implement other bases can affect userlands to irregular BMS sources. To prevent this, I have wanted bibliographies of the BMS extensions.

If I complete #214, suppressing them may be accomplished with custom presets.

PR #221, opened 2025-10-02
add: `BaseType`

MiyakoMeow:


MikuroXina:
I want to establish our position about `#BASE 36` and `#BASE 16`; I'd like to treat it as an illegal command and keep it as to be warned. If you want to accept it somehow, see #223.

And `ObjId` is a not integer but a string of two alpha-numeric characters. I designed it as the string will be uppercase for its case-insensitive equality. But when `#BASE 62` is specified, the conversion is skipped for its case-sensitive equality. The reason why I removed `impl TryFrom<_> for ObjId` is those conversions don't consider its context about case-sensivity. So it's easy to make bugs about `ObjId`'s equality, so I integrated it as a parameter of the conversion method `ObjId::try_from(string, sensitivity)`.

Therefore, I think it's not need to add `BaseType` very much.

MikuroXina:
> This PR is not about `#BASE 16` and `#BASE 36`, though it's the pre-step of implementing them.

As a fact, [there are many implementation about range of object ids](https://hitkey.nekokan.dyndns.info/cmds.htm#WAVXX): `[01-FF]` (original), `[01-FZ]`, `[01-ZZ]` and so on. Due to the differences, it's confusable dangerous to treat them as integers with the base. For example, 85 will be `2D` in `[01-ZZ]` and `55` in `[01-FF]`. So I'd like to not support features treating object id is an integer with the base.

> I think the `ObjId` type should be able to come from `[char; 2]` directly, because the `[char; 2]` info already contains the value itself, then you can also check the value is inside the `BaseType` or not.

Also there is a very important invariant "alpha-numeric characters" for `ObjId`. It is needed for editing/debugging with visibility of BMS entirely. And the original BMS specification implicitly defines its case-insensitivity. So it's needed to wrap with a type to provide its case-insensitive equality.

PR #229, opened 2025-10-14
feat!: controlflowize bms::parse

MiyakoMeow:


MikuroXina (changes requested):
[`ControlFlow`](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html) is "used to tell an operation whether it should exit early or go on as usual." The functions that returns it are in the context such as repetition and needed to tell the callee whether the process exits early or goes on. Both of them are normal process (without any errors). In this view point, `ControlFlow<Result<()>>` looks weird rather than `Result<ControlFlow<()>>`.
Also, methods of implementations for `TokenProcessor` must not tell to break the process. It will interfere creating features using tokens which is already processed in another processor.

PR #230, opened 2025-10-15
feat!: controlflowize `bms::parse::token_processor`

MiyakoMeow:


MiyakoMeow:
Is this necessary?

> The same token can be parsed by more than 1 processor.

If not, I'd like to use ControlFlow like bms::lex::parser.

PR #244, opened 2025-10-28
fix(random): use `expect` for debug panic

MiyakoMeow:


MikuroXina (commented):
`RandomTokenProcessor`'s `state_stack` has a kind of invariant that there is always `ProcessState::Root` which must not be popped. So I want to panic to detect bugs if it happened. Do you have a test case about occurring a panic by this?

PR #249, opened 2025-11-02
feat(chart_processor)!: use id to sign different events, avoid duplicate calculation

MiyakoMeow:


src/chart_process/types.rs
```diff
@@ -198,11 +325,71 @@ impl From<BmpId> for usize {
...
+    pub const fn peek_next(&self) -> ChartEventId {
+        ChartEventId::new(self.next)
+    }
+}
+
 /// Timeline event and position wrapper type.
 ///
 /// Represents an event in chart playback and its position on the timeline.
 #[derive(Debug, Clone)]
 pub struct ChartEventWithPosition {
```
MikuroXina: Maybe we should provide `PartialEq`/`Eq` implementations for `ChartEventWithPosition`.

MikuroXina (approved):

PR #251, opened 2025-11-03
feat(bms)!: use Result<Bms, ParseError> in BmsOutput/ParseOutput, functions return BmsOutput/ParseOutput again

MiyakoMeow:


MiyakoMeow:
> I think Prompter has an extra responsibility about handling soft warnings. It can be thought as a kind of paying attention for users to deal with, but may be violated Interface Segregation Principle.

In my understanding, the responsibility of the Prompter is to detect the occurrence of duplicate ObjIds.
If you would like to deal with soft warnings in this style, why not add an `WarningHandler`?
But the simplest way I'd like is just collecting and returning warnings. The return structure is noticeable for users.
And other modules in this crate like `LexOutput` and `BmsonOutput`, are they worth to change the warning-handling style?

MikuroXina:
To improve the return structure, I think we have another choice that with these steps:

1. Add a new checkpoint type `Checkpoint<'a, 't>(&'a [&'t TokenWithRange<'t>])`
2. Add a new internal context type `ProcessContext<'a, 't, P>` which consists of:
    1. `input: &'a mut &'a [&'t TokenWithRange<'t>]`,
    2. `prompter: P`,
    3. `reported: Vec<ParseWarningWithRange>`,
    4. `fn save<'a>(&'a self) -> Checkpoint<'a, 't>`
    5. `fn restore<'a>(&'a mut self, checkpoint: Checkpoint<'a, 't>)`
    6. `fn prompter(&self) -> &P`,
    7. `fn warn(&mut self, warning: ParseWarningWithRange)`.
3. Change `TokenProcessor::process` to take a `ctx: &mut ProcessContext` instead of `input` and `prompter`.
4. Replace calling `prompter.warn` with calling `ctx.warn`.
5. Change to extract collected warnings from the context data at API endpoints.

How about this plan?

MiyakoMeow:
Implemented in `bms::token_processor`.

MikuroXina (changes requested):

src/bms/model/bmp.rs
```diff
@@ -1,4 +1,4 @@
-//! This module introduces struct [`BmpObjects`], which manages definitions and events of background image animation (BGI).
+//! This module introduces struct [`BmpObjects`], which manages definitions and events of background image and movie.
```
MikuroXina: Note that this module doesn't owe to process movie (`video` module is responsible for that). The "animation" word means that there are objects to switch images to animate them.

src/bms/parse/prompt.rs
```diff
@@ -40,9 +40,18 @@ pub trait Prompter {
...
+/// Blanket implementation to allow borrowing a prompter without cloning.
+impl<T: Prompter + ?Sized> Prompter for &T {
+    fn handle_def_duplication(&self, duplication: DefDuplication) -> DuplicationWorkaround {
+        (*self).handle_def_duplication(duplication)
+    }
+    fn handle_track_duplication(&self, duplication: TrackDuplication) -> DuplicationWorkaround {
+        (*self).handle_track_duplication(duplication)
+    }
+    fn handle_channel_duplication(&self, duplication: ChannelDuplication) -> DuplicationWorkaround {
+        (*self).handle_channel_duplication(duplication)
```
MikuroXina: This implementation should be unnecessary. Because the trait `Prompter` requires only the reference of data which implements it.

src/bms/parse/token_processor/random.rs
```diff
@@ -574,24 +579,41 @@ impl<R: Rng, N: TokenProcessor> RandomTokenProcessor<R, N> {
...
+        let mut view_ctx = ProcessContext {
+            input: &mut tmp,
+            prompter: ctx.prompter(),
+            reported: Vec::new(),
+        };
+        let out = self.next.process(&mut view_ctx)?;
+        // Collect nested warnings locally first, then drop the borrowed context.
+        let nested_reported = core::mem::take(&mut view_ctx.reported);
+        drop(view_ctx);
+        ctx.reported.extend(nested_reported);
```
MikuroXina: I want this recovery and merging warnings to be done with the checkpoint mechanism. The same goes for `impl TokenProcessor for SequentialProcessor`.

MikuroXina (approved):

Issue #243, opened 2025-10-26
New model for random syntax structure


MikuroXina:
I removed AST of random syntax at v0.10.0, so we need to develop a random syntax feature for users who want to deal with them. Currently, [`RandomTokenProcessor`](https://github.com/MikuroXina/bms-rs/blob/d5b5aa0c64956f3ebeefb1d2dec6e40d66a3a9c7/src/bms/parse/token_processor/random.rs) will process tokens with its pushdown automaton. So there are no data structures of random syntax such as AST.

Since I designed an API that can build and modify random syntax below. At the moment, I'm planning only features about `#RANDOM`.

- `struct Random` - It contains inner branches and the maximum number for RNG.
  - `fn new<BS>(max: BigUint, branches: BS) -> Self where BS: IntoIterator<Item=If>`
  - `fn len(&self) -> usize`
  - `fn is_empty(&self) -> bool`
  - `fn at(&self, index: usize) -> Option<&If>`
  - `fn at_mut(&mut self, index: usize) -> Option<&mut If>`
  - `fn push(&mut self, branch: If)`
  - `fn pop(&mut self) -> Option<If>`
  - `fn clear(&mut self)`
  - `fn into_tokens(&
  - `impl Index<usize> for Random { type Output = If; }`
  - `impl IndexMut<usize> for Random`
  - `impl IntoIterator for Random { type Item = If; }`
  - `impl IntoIterator for &Random { type Item = &If; }`
- `struct SetRandom` - It contains the given value and inner branches.
  -  `fn new<BS>(given: BigUint, branches: BS) -> Self where BS: IntoIterator<Item=If>`
  - Others are same as `Random`. It may be better to delegate them to another private inner struct.
- `struct If` - It contains the condition, tokens and next chain item (`IfChain`).
  - `fn new<T>(cond: BigUint, tokens: T) -> Self where T: IntoIterator<Item=Token>`
  - `fn or_else_if(self, cond: BigUint, token: T) -> Self where T: IntoIterator<Item=Token>`
  - `fn or_else(self, tokens: T) -> Self where T: IntoIterator<Item=Token>`
  - `fn at(&self, index: usize) -> Option<&IfChainEntry>`
  - `fn at_mut(&mut self, index: usize) -> Option<&mut IfChainEntry>`
  - `impl Index<usize> for If { type Output = IfChainEntry; }`
  - `impl IndexMut<usize> for If`
[...]

PR #253, opened 2025-11-05
add(bms::model): Random/Switch structure

MiyakoMeow:
#243

### DIFF
--- a/src/bms/model/control_flow.rs
+++ b/src/bms/model/control_flow.rs
@@ -379,22 +379,13 @@ impl<'a> Random<'a> {
     pub fn into_tokens(self) -> Vec<Token<'a>> {
         let mut out = Vec::new();
         match &self.value {
-            ControlFlowValue::GenMax(max) => out.push(Token::Header {
-                name: "RANDOM".into(),
-                args: max.to_string().into(),
-            }),
-            ControlFlowValue::Set(val) => out.push(Token::Header {
-                name: "SETRANDOM".into(),
-                args: val.to_string().into(),
-            }),
+            ControlFlowValue::GenMax(max) => out.push(Token::header("RANDOM", max.to_string())),
+            ControlFlowValue::Set(val) => out.push(Token::header("SETRANDOM", val.to_string())),
         }
 
         self.branches.into_iter().for_each(|branch| {
             // Emit head IF
-            out.push(Token::Header {
-                name: "IF".into(),
-                args: branch.condition.to_string().into(),
-            });
+            out.push(Token::header("IF", branch.condition.to_string()));
             out.extend(
                 branch
                     .head_units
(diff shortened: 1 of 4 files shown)


MikuroXina (changes requested):

src/bms/model/control_flow.rs
```diff
@@ -0,0 +1,724 @@
...
+    /// An `#ELSE` branch with its units and a pointer to the next entry.
+    Else {
+        /// Content units for this branch.
+        units: Vec<TokenUnit<'a>>,
+        /// Pointer to the next chain entry.
+        next: Box<IfChainEntry<'a>>,
+    },
+    /// Terminator for an if-chain.
+    EndIf,
+}
```
MikuroXina: I don't want it to be public, because there should be a invariant that `IfChainEntry::Else`'s `next` field must not be a `IfChainEntry::ElseIf`.
```suggestion
/// One branch node in an if-chain inside a random block.
#[derive(Debug, Clone, PartialEq, Eq)]
enum IfChainEntry<'a> {
    /// An `#ELSEIF <cond>` branch with its units and a pointer to the next entry.
    ElseIf {
        /// Condition value for `#ELSEIF <cond>`.
        cond: BigUint,
        /// Content units for this branch.
        units: Vec<TokenUnit<'a>>,
        /// Pointer to the next chain entry.
        next: Box<IfChainEntry<'a>>,
    },
    /// An `#ELSE` branch with its units and a pointer to the next entry.
    Else {
        /// Content units for this branch.
        units: Vec<TokenUnit<'a>>,
        /// Pointer to the next chain entry.
        next: Box<IfChainEntry<'a>>,
    },
    /// Terminator for an if-chain.
    EndIf,
}
```
Also we need to check and maintain it at `append_to_tail`.

MikuroXina (changes requested):

src/bms/model/control_flow.rs
```diff
@@ -0,0 +1,650 @@
...
+                        }
+                        IfChainEntry::ElseIf { next, .. } => {
+                            cur = next.as_mut();
+                        }
+                        IfChainEntry::Else { .. } => {
+                            // ELSE is terminal; do not append beyond it.
+                            break;
+                        }
+                    }
+                }
```
MikuroXina: It can be done by a recursion.
```suggestion
                next.append_to_tail(entry);
```

MikuroXina (changes requested):

src/bms/model/control_flow.rs
```diff
@@ -0,0 +1,631 @@
...
+    /// `index>=1` refers to the subsequent `ELSEIF`/`ELSE` nodes.
+    /// Returns previous units when the index exists.
+    pub fn set_units_at<U>(&mut self, index: usize, new_units: U) -> Option<Vec<TokenUnit<'a>>>
+    where
+        U: IntoIterator<Item = TokenUnit<'a>>,
+    {
+        if index == 0 {
+            let mut incoming: Vec<TokenUnit<'a>> = new_units.into_iter().collect();
+            std::mem::swap(&mut incoming, &mut self.head_units);
+            Some(incoming)
```
MikuroXina: It can shorten with `replace`:
```suggestion
            let old = std::mem::replace(&mut self.head_units, new_units.into_iter().collect());
            Some(old)
```

src/bms/model/control_flow.rs
```diff
@@ -0,0 +1,631 @@
...
+            out.push(Token::Header {
+                name: "ENDIF".into(),
+                args: "".into(),
+            });
+        });
+
+        out.push(Token::Header {
+            name: "ENDRANDOM".into(),
+            args: "".into(),
+        });
```
MikuroXina: `Token::header` is useful to get a literal token. Also same as it at `Switch::into_tokens`.
```suggestion
                        out.push(Token::header("ELSE", ""));
                        out.extend(units.into_iter().flat_map(TokenUnit::into_tokens));
                        break;
                    }
                    IfChainEntry::EndIf => break,
                }
            }

            // Close the IF-chain
            out.push(Token::header("ENDIF", ""));
        });

        out.push(Token::header("ENDRANDOM", ""));
```

MikuroXina (approved):

PR #257, opened 2025-11-08
fix!: refine `ObjTime` constructor

MiyakoMeow:


src/bms/command/time.rs
```diff
@@ -62,6 +70,16 @@ impl ObjTime {
         }
     }
 
+    /// Create a new time from a track.
+    #[must_use]
+    pub fn new_track(track: u64) -> Self {
+        Self {
+            track: Track(track),
```
MikuroXina: `ObjTime` expresses a point of time on the BMS score, not a track. So I'd like to name it like "start of the track" for what you want to do that.
```suggestion
    pub fn start_of(track: Track) -> Self {
        Self {
            track: track,
```

MikuroXina (changes requested):

PR #258, opened 2025-11-12
fix(test): use doc test than ignore

MiyakoMeow:


src/bms/model/notes.rs
```diff
@@ -215,8 +229,12 @@ impl Notes {
     /// to the Notes object.
     ///
     /// To filter out dangling objects, use:
-    /// ```rust,ignore
+    /// ```rust,no_run
+    /// # use bms_rs::bms::prelude::ObjTime;
+    /// # let notes = bms_rs::bms::model::Notes::default();
+    /// # let time_span = ObjTime::new(1, 0, 4).unwrap()..ObjTime::new(2, 0, 4).unwrap();
```
MikuroXina: I think that we should show how to get `time_span`.
```suggestion
    /// let time_span = ObjTime::new(1, 0, 4).unwrap()..ObjTime::new(2, 0, 4).unwrap();
```

PR #259, opened 2025-11-12
feat(chart_processor)!: add activate_time

MiyakoMeow:


src/chart_process/bms_processor.rs
```diff
@@ -612,6 +696,136 @@ impl BmsProcessor {
         )
     }
 
+    /// Precompute absolute activate_time for all events based on BPM segmentation and Stops.
+    fn precompute_activate_times(&mut self) {
```
MikuroXina: I think that it is hard to guarantee that `all_events` is holding the invariants. Also the precomputation process looks too complex. How about to split it into another private struct?

src/chart_process/bms_processor.rs
```diff
@@ -61,7 +65,11 @@ impl BmsProcessor {
         let default_visible_y_length =
             compute_default_visible_y_length(base_bpm.clone(), reaction_time);
 
-        let all_events = Self::precompute_all_events::<T>(&bms);
+        let pre_index = {
+            let bms: &Bms = &bms;
+            BmsPrecomputer::precompute_all_events::<T>(bms)
+        };
```
MikuroXina: Why don't you simply do this?
```suggestion
        let pre_index = BmsPrecomputer::precompute_all_events::<T>(&bms);
```

src/chart_process/bms_processor.rs
```diff
@@ -759,7 +542,8 @@ impl ChartProcessor for BmsProcessor {
         let visible_window_y = self.visible_window_y();
         let scroll_factor = self.current_scroll.clone();
 
-        self.preloaded_events.iter().map(move |event_with_pos| {
+        let preloaded = self.preloaded_events.clone();
+        preloaded.into_iter().map(move |event_with_pos| {
```
MikuroXina: It's unnecessary to clone the vec.
```suggestion
        self.preloaded_events.iter().map(move |event_with_pos| {
```

src/chart_process/bms_processor.rs
```diff
@@ -787,3 +564,433 @@ enum FlowEvent {
...
+            for argb_obj in argb_changes.values() {
+                let y = BmsProcessor::y_of_time_static(
+                    bms,
+                    argb_obj.time,
+                    &bms.speed.speed_factor_changes,
+                );
+                let argb = ((argb_obj.argb.alpha as u32) << 24)
+                    | ((argb_obj.argb.red as u32) << 16)
+                    | ((argb_obj.argb.green as u32) << 8)
+                    | (argb_obj.argb.blue as u32);
```
MikuroXina: This conversion should be extracted as a method of `Argb` into `bms::command::graphics` module.

MikuroXina (changes requested):

MikuroXina (approved):

PR #263, opened 2025-11-19
feat(bms)!: use TokenModifier than use_relaxed

MiyakoMeow:


src/bms/parse/token_processor.rs
```diff
@@ -293,6 +294,80 @@ pub(crate) fn full_preset<T: KeyLayoutMapper, R: Rng>(
...
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
+/// A relaxed modifier that normalizes common BMS typos and formats,
+/// delegating to `rewrite_relaxed_tokens`.
+pub struct DefaultTokenRelaxer;
+
+impl TokenModifier for DefaultTokenRelaxer {
+    fn modify(&self, tokens: &mut TokenStream<'_>) {
+        relax_tokens_default(tokens);
+    }
+}
```
MikuroXina: Alike as `TokenProcessor` for `SequentialProcessor`, you can add a sequential join type for two of `TokenModifier`s. If you do this, `ParseConfig` can avoid to use `Vec<Box<dyn TokenModifier>>` and available the static dispatch with type parameters.

PR #266, opened 2025-11-26
feat(control_flow)!: use `bms::model::control_flow` to parse control flows

MiyakoMeow:


src/bms/model/control_flow/activate.rs
MikuroXina: This new model should not depend on the existing one which made for users to build randomization control flow tokens. Because they are very different conceptually. For example, to evaluate a built random block, it's required for an unit of token to contain a range of the source, and to prepare a random number generator.

So If you want to do that, I recommend you extracting this module from `bms::model::control_flow` into another one such as `src/bms/parse/token_processor/random/activate.rs` and so on. And then it's better to define a new control flow model wrapped in `SourceRangeMixin` over there.

MikuroXina: The important difference between before and now (you wanted) is this feature aims to model abilities of choices of control flow paths (right?). To accomplish your goal, the way of AST and parser is not absolutely necessary. It will be ok to change the `random` token processor as extracting all the randomization branches to keep them in a model data structure. I think it's too complex to realize it by adding the AST, parser and interpreter.

MiyakoMeow: OK, I accept your proposal.
Please note here if you finish the modification, and tell me what can I do.

PR #271, opened 2025-12-05
feat(chart_processor)!: use PointerSpeed/VisibleRangePerBpm

MiyakoMeow:


src/chart_process/types.rs
```diff
@@ -55,6 +55,109 @@ impl From<Decimal> for BaseBpm {
     }
 }
 
+/// Pointer speed per BPM, representing the movement speed of the playhead in Y units per second per BPM.
+/// Formula: (Y/sec)/bpm
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PointerSpeed(Decimal);
```
MikuroXina: What is the "pointer"? Does it mean the playhead? If it is, you should use `PlayheadSpeed`.

Also if you anticipated moving the playhead on the score, they are opposite. Modeling the game play, I think there are the items: the fixed playhead and moving notes. It's more natural because falling notes can accelarate or going back.

MiyakoMeow: BMS has no build-in "accelarate" or "notes going back" function, but these can be simulated by `#BPM`, `#SPEED` and `#SCROLL`.
The `Y` is always going ahead in the process.
Also, the movement of notes is now calculated by `ChartProcessor` implements.
So I think the logic now is almost natural for BMS/BMSON.

MiyakoMeow: For `PlayheadSpeed`:
The formula is: `(Y/sec)/bpm`. During the process, without modification, the value is `PlayheadSpeed::standard`.
Now I provided a new `ControlEvent` member to modify `PlayheadSpeed`, the only function is adjusting the playback speed.
...
It seems just providing a more obvious function.
Users can just pass the modified `SystemTime` value when calling `ChartProcessor::update` to do the same thing.

src/chart_process/bmson_processor.rs
```diff
@@ -3,19 +3,20 @@
...
-    types::{
-        AllEventsIndex, BaseBpm, BmpId, ChartEventIdGenerator, DisplayRatio, WavId, YCoordinate,
-    },
+    ChartEvent, ChartProcessor, ControlEvent, PlayheadEvent, VisibleChartEvent, VisibleRangePerBpm,
+    types::{AllEventsIndex, BmpId, ChartEventIdGenerator, DisplayRatio, WavId, YCoordinate},
 };
 use num::{One, ToPrimitive, Zero};
 
+/// Static OnceLock for Decimal::one() to avoid repeated allocations
+static DECIMAL_ONE: OnceLock<Decimal> = OnceLock::new();
```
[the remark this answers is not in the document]
MikuroXina: Um, did you say about `ControlEvent::SetPlaybackRatio`? My suggestion is unrelated to change other values or to change `current_speed`'s return type. What's your consideration?

PR #283, opened 2025-12-26
fix: use `AsRef` for new-type struct, and fix unneed clone/ref

MiyakoMeow:


src/bms/command/minor_command.rs
```diff
@@ -10,6 +10,18 @@ use crate::bms::command::time::ObjTime;
...
+    fn as_ref(&self) -> &i64 {
+        &self.0
+    }
+}
+
+impl From<ExWavPan> for i64 {
+    fn from(value: ExWavPan) -> Self {
+        value.0
+    }
+}
```
MikuroXina: `ExWavPan` and `ExWavVolume` should not impl `From` conversion from the internal type because they have own additional invariant about the contained.

MikuroXina (changes requested):

src/bms/command/minor_command.rs
```diff
@@ -18,12 +24,6 @@ impl ExWavPan {
         (-10000..=10000).contains(&value).then_some(Self(value))
     }
 
-    /// Returns the underlying value.
-    #[must_use]
-    pub const fn value(self) -> i64 {
-        self.0
-    }
```
MikuroXina: I think `value` methods are still useful. It can extract the contained by one shot.

src/chart_process/bmson_processor.rs
```diff
@@ -393,7 +393,7 @@ impl ChartProcessor for BmsonProcessor {
             // Calculate display ratio: (event_y - current_y) / visible_window_y * scroll_factor
             // Note: scroll can be non-zero positive or negative values
             let display_ratio_value = if visible_window_y > YCoordinate::zero() {
-                (&(event_y - current_y) / &visible_window_y).value() * scroll_factor
+                Decimal::from(&(event_y - current_y) / &visible_window_y) * scroll_factor.clone()
```
MikuroXina:
```suggestion
                &Decimal::from(&(event_y - current_y) / &visible_window_y) * scroll_factor
```
This `clone` can be reduced by aligning ref.

MikuroXina (changes requested):

MikuroXina (approved):

PR #285, opened 2025-12-27
fix: add lints

MiyakoMeow:


MikuroXina (changes requested):

src/bms/command/channel/mapper.rs
```diff
@@ -33,14 +32,15 @@ fn key_layout_beat_to_channel_id(beat: KeyLayoutBeat) -> NoteChannelId {
         _ => '1', // Default fallback
     };
 
-    NoteChannelId::try_from([first_char as u8, second_char as u8]).unwrap()
+    let Ok(channel_id) = NoteChannelId::try_from([first_char as u8, second_char as u8]) else {
+        panic!("generated note channel id should be valid");
+    };
+    channel_id
```
MikuroXina: I think those of panicking with message of invariant are suitable to use `expect` over `panic!`. I don't want to maintain about what format to print as possible.

src/bms/command/minor_command.rs
```diff
@@ -143,7 +143,7 @@ pub struct StpEvent {
     pub duration: Duration,
 }
 
-/// MacBeat `#WAVCMD` event.
+/// `MacBeat` `#WAVCMD` event.
```
MikuroXina: MacBeat is a proper nourn, software name. It should not markup as an inline code.

src/bms/parse/token_processor/bmp.rs
```diff
@@ -60,35 +60,36 @@ impl TokenProcessor for BmpProcessor {
...
         })?;
         Ok(objects)
     }
 }
 
 impl BmpProcessor {
+    fn bga_layer(channel: Channel) -> Result<BgaLayer> {
+        BgaLayer::from_channel(channel).ok_or(ParseWarning::SyntaxError(format!(
+            "invalid channel for BgaLayer: {channel:?}",
+        )))
```
MikuroXina:
```suggestion
        BgaLayer::from_channel(channel).ok_or_else(|| {
            ParseWarning::SyntaxError(format!("invalid channel for BgaLayer: {channel:?}",))
        })
```
You should use `ok_or_else` to avoid unnecessary allocation by `format!`.

src/bmson/fin_f64.rs
```diff
@@ -77,6 +77,6 @@ impl FinF64 {
 impl TryFromFloatError {
     /// Returns the contained error payload.
     pub const fn value(self) {
-        self.0
+        let _ = self;
```
MikuroXina: This is a mistake of the definition, and should be fixed as:
```rust
    pub const fn value(self) -> f64 {
        self.0
    }
```

MikuroXina (changes requested):

src/bmson/bmson_to_bms.rs
```diff
@@ -332,7 +331,6 @@ fn convert_lane_to_key_side(lane: Option<NonZeroU8>) -> (Key, PlayerSide) {
 
     // Convert lane to key
     let key = match adjusted_lane {
-        1 => Key::Key(1),
```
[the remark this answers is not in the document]
MikuroXina: Oh, the fallback arm is occurring a warning. It is intentional and should not be removed.

MikuroXina (changes requested):

MikuroXina (changes requested):

MikuroXina (approved):

PR #293, opened 2026-01-02
fix(chart_process): use round() to avoid misalignment

MiyakoMeow:


src/chart_process/bms_processor.rs
```diff
@@ -539,15 +539,15 @@ fn precompute_activate_times(
...
-            total_nanos = total_nanos.saturating_add(delta_nanos_f64.round() as u64);
-        }
+        let delta_y = curr.clone() - prev.clone();
+        let delta_y_value = delta_y.value();
+        let delta_nanos = if cur_bpm > Decimal::zero() {
+            let numerator = delta_y_value * Decimal::from(240u64) * Decimal::from(NANOS_PER_SECOND);
+            (numerator / cur_bpm).round().to_u64().unwrap_or(0)
+        } else {
+            0
+        };
```
MikuroXina: Since you are unwrapping the result, it's not needed to check whether `cur_bpm` is positive:
```suggestion
        let delta_nanos = (delta_y_value * Decimal::from(240u64) * Decimal::from(NANOS_PER_SECOND)
            / cur_bpm)
            .round()
            .to_u64()
            .unwrap_or(0);
```
Same on `bmson_processor`.

MikuroXina (changes requested):

MikuroXina (approved):

PR #307, opened 2026-01-13
add(example): microquad_player

MiyakoMeow:


MikuroXina (approved):
It looks good, but `player.visible_events()` is not returning long notes which are passed its start. I'll make it an issue later.

Issue #299, opened 2026-01-10
Allow `#XXX02:0.0` situation in BMS


MiyakoMeow:
Now Bms parser will directly return Error when meeting #XXX02:0.0.
I think this is reasonable for just skipping track numbers.

And If this is allowed, as #279 , there should be tests for `BmsProcessor` and `ChartPlayer`.

PR #311, opened 2026-01-18
fix(bms): allow #XXX02:0.0

MiyakoMeow:
#299

### DIFF
--- a/src/chart_process/bms_processor.rs
+++ b/src/chart_process/bms_processor.rs
@@ -65,8 +65,7 @@ impl BmsProcessor {
             .map(|(obj_id, bmp)| (BmpId::from(obj_id.as_u16() as usize), bmp.file.clone()))
             .collect();
 
-        let (all_events, flow_events_by_y) =
-            AllEventsIndex::precompute_all_events::<T>(bms, &y_memo);
+        let all_events = AllEventsIndex::precompute_all_events::<T>(bms, &y_memo);
 
         // Precompute activate times
         let all_events = precompute_activate_times(bms, &all_events, &y_memo);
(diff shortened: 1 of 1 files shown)


src/chart_process/bms_processor.rs
```diff
@@ -200,195 +195,337 @@ impl YMemo {
...
     /// Note: Speed effects are calculated into event positions during initialization, ensuring event trigger times remain unchanged
+    /// Returns (events index, flow events by y)
     #[must_use]
-    pub fn precompute_all_events<T: KeyLayoutMapper>(bms: &Bms, y_memo: &YMemo) -> Self {
+    pub fn precompute_all_events<T: KeyLayoutMapper>(
+        bms: &Bms,
+        y_memo: &YMemo,
+    ) -> (Self, BTreeMap<YCoordinate, Vec<FlowEvent>>) {
         let mut events_map: BTreeMap<YCoordinate, Vec<PlayheadEvent>> = BTreeMap::new();
+        let mut flow_events_by_y: BTreeMap<YCoordinate, Vec<FlowEvent>> = BTreeMap::new();
```
MikuroXina: How about to add `flow_events_by_y` as the `YMemo`'s field? It would be more intutive.

src/chart_process/bms_processor.rs
```diff
@@ -421,65 +556,77 @@ pub fn precompute_activate_times(
...
+            if *change_y <= curr {
+                cur_bpm = bpm_map
+                    .range(..=change_y)
+                    .next_back()
+                    .map(|(_, b)| b.clone())
+                    .unwrap_or_else(|| init_bpm.clone());
+                bpm_iter.next();
+            }
+        }
+
```
MikuroXina: If you collect `bpm_changes` as `BTreeMap<YCoordinate, Decimal>`, the code could be written like this:

```suggestion
        if let Some((_, bpm)) = bpm_changes.range(..=&curr).next_back() {
            cur_bpm = bpm.clone();
        }
```

MikuroXina (changes requested):

Issue #309, opened 2026-01-13
Unify the evaluation criterion of `chart_process`'s time precision


MikuroXina:
Test cases for `chart_process` processes time seems inconsistent. So then I'd like to unify the evaluation criterion of them as:

- Absolute error of the result `(expected - actual).abs()` must be less than 1 microseconds.

Because if accumulated the calculation errors by thousands of events, it corrupts only a few milliseconds.

PR #313, opened 2026-01-18
fix(test): time assert

MiyakoMeow:
#309

### DIFF
--- a/tests/chart_process/mod.rs
+++ b/tests/chart_process/mod.rs
@@ -10,7 +10,7 @@ pub mod edge_cases;
 pub mod update_consistency;
 
 /// Unified time precision evaluation constant: 1 microsecond (unit: seconds)
-pub(super) const MICROSECOND_EPSILON: f64 = 1e-6;
+const MICROSECOND_EPSILON: f64 = 1e-6;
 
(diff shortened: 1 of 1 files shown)


tests/chart_process/mod.rs
```diff
@@ -1,6 +1,38 @@
...
+//! - All time-related assertions must have absolute error less than 1 microsecond (0.000001s)
+//! - Use [`assert_time_close`] function for unified time precision evaluation
 
 pub mod bms_processor;
 pub mod bmson_processor;
 pub mod edge_cases;
 pub mod update_consistency;
+
+/// Unified time precision evaluation constant: 1 microsecond (unit: seconds)
+pub(super) const MICROSECOND_EPSILON: f64 = 1e-6;
```
MikuroXina:
```suggestion
const MICROSECOND_EPSILON: f64 = 1e-6;
```
You don't need `pub(super)` to export it for children module. `pub(super)` means that the item is exported against its super (`tests/mod.rs`) module.

MikuroXina (changes requested):

MikuroXina (approved):

PR #318, opened 2026-01-19
doc(chart_process): tidy formulas

MiyakoMeow:


src/chart_process.rs
```diff
@@ -1,10 +1,67 @@
 //! Chart Processor
 //!
-//! Unified Y coordinate definition:
+//! # Y Coordinate Definition
```
MikuroXina: I'd like to make the title `Chart Processor` into a level 1 heading and make `Y Coordinate Definition` into a level 2 heading.
```md
# Chart Processor

## Y Coordinate Definition
```

PR #331, opened 2026-02-04
refactor(bms)!: rm fraction, use StringValue for `FromStr` values

### LINKED ISSUE #289: Is `Decimal` (`fraction::BigDecimal`) necessary?
I have added the `fraction::BigDecimal` usage several months ago.
But I found these problems:
- I've not seen a `bms`/`bmson` file that must define the value exceeding `u64` range.
- Extra code for type converting.
- Unavoidable heap clones.
- A little confusing usage for users.

Maybe it's better to remove the usage of `fraction::BigDecimal`.

Tips: for values restriction like finite/positive f64, [`strict-num`](https://crates.io/crates/strict-num) may be useful.

MiyakoMeow:
Resolves #289

### DIFF
--- a/src/chart_process/player.rs
+++ b/src/chart_process/player.rs
@@ -140,7 +140,7 @@ impl ChartPlayer {
         let cur_y = self.playback_state.progressed_y;
 
         // Calculate preload range: current y + visible y range
-        let visible_y_length = self.visible_window_y(&speed);
+        let visible_y_length = self.visible_window_y(speed);
         let preload_end_y = cur_y + visible_y_length;
 
         use std::ops::Bound::{Excluded, Included};
@@ -148,7 +148,7 @@ impl ChartPlayer {
         // Collect events triggered at current moment
         let mut triggered_events = self.events_in_y_range((Excluded(&prev_y), Included(&cur_y)));
 
-        self.update_preloaded_events(&FinF64::new(preload_end_y.as_f64()).unwrap_or(MAX_FIN_F64));
+        self.update_preloaded_events(FinF64::new(preload_end_y.as_f64()).unwrap_or(MAX_FIN_F64));
 
         // Apply Speed changes
         for event in &triggered_events {
@@ -202,8 +202,8 @@ impl ChartPlayer {
 
     /// Gets the current visibility range.
     #[must_use]
-    pub const fn visibility_range(&self) -> &(Bound<FinF64>, Bound<FinF64>) {
-        &self.visibility_range
+    pub const fn visibility_range(&self) -> (Bound<FinF64>, Bound<FinF64>) {
+        self.visibility_range
     }
 
     /// Set playback ratio.
@@ -263,7 +263,7 @@ impl ChartPlayer {
         &mut self,
     ) -> Vec<(PlayheadEvent, std::ops::RangeInclusive<DisplayRatio>)> {
         let current_y = &self.playback_state.progressed_y;
-        let visible_window_y = self.visible_window_y(&self.playback_state.current_speed);
+        let visible_window_y = self.visible_window_y(self.playback_state.current_speed);
         let scroll_factor = &self.playback_state.current_scroll;
 
         let view_start = *current_y;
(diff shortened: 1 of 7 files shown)


src/bms/command/string_value.rs
```diff
@@ -0,0 +1,139 @@
...
+            raw: value.to_string(),
+            value: Ok(value),
+        }
+    }
+
+    /// Creates a `StringValue` from a raw string.
+    #[must_use]
+    pub fn new(raw: String) -> Self {
+        let value = T::from_str(&raw);
+        Self { raw, value }
```
MikuroXina: How about to make `raw` an `impl Into<String>`? It will help us reducing explicit conversions.
```suggestion
    pub fn new(raw: impl Into<String>) -> Self {
        let raw = raw.into();
        let value = T::from_str(&raw);
        Self { raw, value }
```

src/bms/model/stop.rs
```diff
@@ -32,7 +35,8 @@ impl StopObjects {
         self.stops
             .entry(stop.time)
             .and_modify(|existing| {
-                existing.duration = &existing.duration + &stop.duration;
+                let sum = existing.duration.as_f64() + stop.duration.as_f64();
+                existing.duration = FinF64::new(sum).unwrap_or(ZERO_FIN);
             })
             .or_insert_with(|| stop);
```
MikuroXina: It should be discouraged to compose `StringValue`s. I'd like to change composing stop objects into storing `stop` as is. Because I want `StopObjects` to express syntactic structure. not a format which can play directly easily.

MikuroXina (changes requested):

src/bms/parse/prompt.rs
```diff
@@ -60,9 +62,9 @@ pub enum DefDuplication<'a> {
         /// Duplicated BPM object id.
         id: ObjId,
         /// Existing definition.
-        older: Decimal,
+        older: &'a core::result::Result<PositiveF64, <PositiveF64 as FromStr>::Err>,
```
MikuroXina: Why do use `Result` of conversion, not a `StringValue`? It seems that this doesn't contain information of the original represented string.

MikuroXina (changes requested):

src/chart_process/processor/bms.rs
```diff
@@ -7,17 +7,22 @@ use std::{
...
 };
-use crate::chart_process::{ChartEvent, FlowEvent, PlayheadEvent, TimeSpan, YCoordinate};
-
+use crate::chart_process::{ChartEvent, FlowEvent, PlayheadEvent, TimeSpan};
+use strict_num_extended::NonNegativeF64;
 const NANOS_PER_SECOND: u64 = 1_000_000_000;
 
+const ZERO_FIN: FinF64 = FinF64::new_const(0.0);
+const DEFAULT_BPM_FIN: PositiveF64 = PositiveF64::new_const(120.0);
+const DEFAULT_SPEED_POSITIVE: PositiveF64 = PositiveF64::new_const(1.0);
```
MikuroXina:
```suggestion
const DEFAULT_BPM: PositiveF64 = PositiveF64::new_const(120.0);
const DEFAULT_SPEED: PositiveF64 = PositiveF64::new_const(1.0);
```
I think constant names should not have a prefix about its type.

MikuroXina (changes requested):

src/chart_process.rs
```diff
@@ -288,131 +287,9 @@ impl std::hash::Hash for PlayheadEvent {
...
+    Speed(PositiveF64),
     /// Scroll factor change event.
-    Scroll(Decimal),
-}
-
-/// Y coordinate wrapper type, using arbitrary precision decimal numbers.
-///
-/// Unified y unit description: In default 4/4 time, one measure equals 1; BMS uses `#SECLEN` for linear conversion, BMSON normalizes via `pulses / (4*resolution)` to measure units.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct YCoordinate(pub Decimal);
```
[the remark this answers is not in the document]
MikuroXina: They are different circumstantially, but I think `YCoordinate` and `BaseBpm` have more context and constraint about what we're doing. Also we might introduce another type of concept that uses `NonNegative`/`Positive` f64 later.

src/chart_process/player.rs
MikuroXina: It seems that you are using `expect` to assert calculation errors, but I think that most of multiplication among large values can occur overflowing to infinity easily. Do you have any idea about handling calculation overflow and so on?

src/chart_process/player.rs
```diff
@@ -532,20 +579,28 @@ impl ChartPlayer {
...
+        let note_min = if ratio_start < ratio_end {
+            ratio_start
+        } else {
+            ratio_end
+        };
+        let note_max = if ratio_start > ratio_end {
+            ratio_start
+        } else {
+            ratio_end
+        };
```
MikuroXina: These two branches can be written as:
```suggestion
        let (note_min, note_max) = if ratio_start < ratio_end {
            (ratio_start, ratio_end)
        } else {
            (ratio_end, ratio_start)
        };
```

PR #335, opened 2026-02-18
feat(bms)!: remove BigUInt usage

brightmeowso:


MikuroXina (changes requested):
As I said before, I think `BitgUint` can't be removed from the model of syntactic items. Do you have any plans?

PR #357, opened 2026-05-01
chore: use #[expect] instead of #[allow] for clippy lints

MiyakoMeow:


src/bms/command/minor_command.rs
```diff
@@ -143,7 +143,7 @@ pub struct StpEvent {
     pub duration: Duration,
 }
 
-#[allow(clippy::doc_markdown)]
+#[expect(clippy::doc_markdown)]
```
MikuroXina: Unlike the case of `panic`, this lint is unintended for our doc comment. So it should be preferred `allow` over `expect`.

MikuroXina (changes requested):

MikuroXina (approved):

PR #360, opened 2026-05-02
refactor(bms)!: move key mapping to chart module

MiyakoMeow:
## Summary

Move key mapping related code from `bms` layer to `chart` layer to decouple parsing from key layout assumptions.

## Changes

- Move `KeyLayoutBeat` and related types from `bms::command::channel::mapper` to `chart::key_mapping`
- Move `check_playing.rs` and `validity.rs` from `bms/parse/` to `chart/key_mapping/check.rs`
- Replace `bms.check_playing::<T>()` method with standalone `check_bms_playing(&bms)` function
- Simplify `BmsWarning` enum by removing `PlayingWarning` and `PlayingError` variants (now handled separately)

## Breaking Changes

- `BmsWarning` enum reduced from 4 to 2 variants
- `parse_bms` no longer includes playing check results; use `check_bms_playing(&bms)` explicitly

## Motivation

The BMS model should be key-layout agnostic. Playing checks and validity validation should not depend on a specific key layout during parsing. This refactor prepares for supporting multiple key layouts without modifying the BMS data model.


MikuroXina:
Hmm, I don't think that "BMS model should be key-layout agnostic". The BMS sources made expected to play its specific key layout. It doens't seem that supporting mutiple key layout (not only logical ones) is reponsible of our BMS model. I think remapping input keys for actual machines is owed to BMS players.

MiyakoMeow:
That's the direction. I'm moving the `KeyMapping` part from model to chart/playing part in this library.

MikuroXina:
I don't know what the direction you saying is. `KeyLayoutBeat` and related types are used to represent a format of logical key layout as static types. Also they implement `KeyLayoutMapper` to get (de facto standard) note channel id, not for playing (at least I thought).

If we were to refactor it, organize the concepts to rename them and improve docs.

MiyakoMeow:
I drafted this because it make the structures not clear.
See #364 
I'd like to do this in another PR.

PR #368, opened 2026-05-08
refactor: unify parsing API under Process trait, hide processor internals

MiyakoMeow:
- Add generic type param L to Process trait for key layout mapping
- Make BmsProcessor and BmsonProcessor crate-private
- Remove TryFrom<Bms/Bmson> for Chart (superseded by Process trait)
- All callers use Process::process() uniformly

### DIFF
--- a/src/bms/process.rs
+++ b/src/bms/process.rs
@@ -2,7 +2,6 @@
 
 use std::{
     collections::{BTreeMap, HashMap},
-    convert::TryFrom,
     path::PathBuf,
 };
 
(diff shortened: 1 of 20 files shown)


src/bms/process.rs
```diff
@@ -200,6 +200,14 @@ impl BmsProcessor {
     }
 }
 
+impl<L: KeyLayoutMapper> Process<L> for Bms {
+    type Error = PlayingError;
+
+    fn process(&self) -> Result<Chart, Self::Error> {
+        BmsProcessor::parse::<L>(self)
+    }
+}
```
MikuroXina, line 209: Hmm, it seems that there is no choice for it. But For further improvement, I thought `Bms` should have the type parameter `L`, not `Process`. `L` will help us to represent that a `Bms` instance has only notes over the layout `L`.

MikuroXina (approved):

PR #369, opened 2026-05-31
chore: ignore .worktrees/

ghost:


MikuroXina (changes requested):
There is no agreement about using git worktree in this project. If you want to deal with git worktree, you should ignore them in your personal global Git configuration file.

PR #374, opened 2026-06-03
feat(judge): add JudgeLevel::VeryEasy for RANK 4

ghost:
Adds  variant for  (VERY EASY, ±25ms, beatoraja/nanasi extension). Previously value 4 fell into  with a note 'decided not to implement'.

Changes:
- New  variant in  enum
-  maps 4 → 
- BMSON conversion updated to match on  instead of


ghost:
Emm...
The original PR is missing because my original account has been suspended for no reason.

MikuroXina (changes requested):
As I remember, the judge level 4 (very easy) is treated as an extension introduced by nanasigroove. Precisely, it is not unimplemented, but covered by `OtherInt` variant.
Interpretion of `JudgeLevel` is mostly implementation-dependent. I wouldn't like to add variants as possible for recommending developers to provide homogeneous user experience.

PR #380, opened 2026-08-08
fix(player): trigger events at y=0 on the first update

### LINKED ISSUE #379: ChartPlayer.update first frame interval excludes events at y = 0 (e.g., BGM on line #00001)
I'm using this library to develop a BMS player, and I encountered the problem that some beatmaps' bgm missing (I tested Random(sltr&sobrem) 7k normal) then AI firmly tell me that "ChartPlayer.update first frame interval excludes events at y = 0 (e.g., BGM on line \#00001)" thus the first bgm event been skipped.

So I want to ensure that if this is really the library's problem or it's just a intentional design.

This is how AI fix it:

Image

naroxeno:
fix #379


MikuroXina (changes requested):
Thanks for contribution. To confirm the validity of your fix, could you add an unit test into `tests` module at the bottom of this module?

MikuroXina (commented):
Hmm, it seems that you have removed the test, but I want to accept it without non-English (Chinese) comments.

